PR #43, opened 2024-03-20
remove unnecessary RUN commands

lpmi-13:
`MKDIR` creates the directory if it doesn't exist, so we can remove these extra `RUN` layers.

https://docs.docker.com/reference/dockerfile/#workdir


amorey:
Thanks! Good to know.

PR #44, opened 2024-03-21
Update README.md with ingress doc

nodesocket:
Add doc for using ingress with Helm chart for access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ helm install kubetail kubetail/kubetail \
 
 ## Access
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +102,27 @@ kubectl auth-proxy -n kubetail http://kubetail.svc
...
+
+Set the following in your `values.yaml` replacing where applicable:
+
+```yaml
+ingress:
+  enabled: true
+  annotations:
+    traefik.ingress.kubernetes.io/router.entrypoints: websecure
+    traefik.ingress.kubernetes.io/router.tls: "true"
+    traefik.ingress.kubernetes.io/router.tls.certresolver: le
```
ethanm12343, line 115: I would remove these, not everyone uses traefik, may confuse inexperienced users :)

rxinui:
@amorey should we close it?

Issue #13, opened 2024-02-20
Support for search / highlight


alexcherka:
Would be really nice if you could filter console lines via search. Or even just highlight a desired key. I know this may not be the intention of this tool but when filtering back in time, would be great to search for specific data if needed!

amorey:
Thanks for the suggestion! I definitely want to add search/highlight functionality to kubetail. It will take some time to implement but now's a good time to start specing it out. Is there another tool that you think has implemented log search particularly well so far?

alexcherka:
As far as web k8s logs viewers go, I don't think there is much out there honestly... I know for IDE's I use K8s Lens that when you view a pod's logs it allows for searching via highlight and then jump around to results. 
![image](https://github.com/kubetail-org/kubetail/assets/9470827/de41dc63-a720-4261-b67f-f79b79b4a90e)

I think this is the best option depending on how you render your front end results. I assume there is some JS packages that allow for search/highlight content on page.

Found something that seems to cover it pretty well.
https://markjs.io/
Specifically the example [Search bar with jump to matches](https://jsfiddle.net/julmot/973gdh8g/)

amorey:
Thanks! I'll take a look at K8S Lens and mark.js.

nefarius:
I think in this case client-side search and indexing makes sense; no load increase on the backend and the client already has all the searchable data anyway so why not make use of it.

alexcherka:
I agree with @nefarius, The front end data is loaded in the first place regardless and then filtering can be done client/DOM level and in theory no extra heavy lifting on the backend or really front end either. I think a highlighting approach is best and just have it jump around to matches. Similar to vscode in file search or any other ide as an example

amorey:
Currently, the frontend only stores new events so the full log files aren't directly accessible locally. We could change this behavior and download the full files in the background and that would work well for small/medium files but it could be problematic for large amounts of data (and data egress costs could also be an issue down the road for power users). I'm still curious to test out client side search though because it might be useful for small amounts of data or in other contexts.

sawan22071995:
Can you tell me how we will generated token when we use tokenauth.yaml. you didn't mention any commads or steps to get the token value to login in server.

amorey:
[...]

PR #172, opened 2025-03-17
Filter and highlight log keyword

cnef:
Filter and highlight log keyword

snapshot


amorey:
@cnef Here's an update related to grep/search: https://github.com/kubetail-org/kubetail/issues/13#issuecomment-2750136772

Issue #184, opened 2025-04-07
Add support for multiple kubeconfig files


menardorama:
Hi,

It would be great if kubetail could handle kubeconfig differently.

For instance I use https://github.com/sbstp/kubie to manage my multiple context; so I have several context files in my $HOME/.kube and no $HOME/.kube/config file.

kubie copy the selected config file in /tmp and set a KUBECONFIG env variable.

Do you think you can update your tool to handle multiple config file in the $HOME/.kube directory ?

Thanks anyway, this tool look promising

amorey:
No problem! As long as the config files use the expected formating, handling multiple files shouldn't be difficult. This is is the code that will have to be modified:
https://github.com/kubetail-org/kubetail/blob/main/modules/shared/k8shelpers/kube-config-watcher.go

I marked this as "good first issue" in case somebody wants to take a stab at it before I do.

sureshopsgeek:
@amorey I would like to work on this.

amorey:
Sounds great!!! I assigned the ticket to you. I'm happy to help get you up-to-speed about the code. If you want to chat about this or just chat in general, join us in Discord:
https://discord.gg/CmsmWAVkvX

amorey:
@menardorama Just a quick update...

@sureshopsgeek has a [PR](https://github.com/kubetail-org/kubetail/pull/192) ready to go but first I wanted to understand your situation better. Here are some questions:

* Do you want kubetail to automatically merge the config files in `~/.kube`or are you looking for a way to specify the config file to use (e.g. via a `--kubeconfig` flag)?
* Are there situations where different config files wil use use the same cluster/user/context names?

menardorama:
Hi

We use dedicated config files on purpose and having them merged is not a valid use case for us...

It makes easier to distribute them

sureshopsgeek:
@menardorama We don't want you to merge them , We merge all kubeconfigs under ~/.kube in kubetail runtime and present you all cluster context. that satisfies your use case?

amorey:
@menardorama Apologies for the confusion. By "merge" I meant virtually, not literally. We're wondering if you want to be able to pick from any of the contexts present across the files or if you just want to specify which config file to use when you start the dashboard (e.g. `kubetail serve --kube-config .kube/config-1.yaml`).

kabakakao:
@amorey @rxinui: I would also love to see, that the KUBECONFIG Environment Variable will/can be used:
[...]

PR #192, opened 2025-04-13
support multiple kubecofig files

sureshopsgeek:
ref: https://github.com/kubetail-org/kubetail/issues/184


amorey:
Does this handle fsnotify "MODIFY" events?

sureshopsgeek:
i added fsnotify.write now. Since my editor is not doing in place replace , it did not capture this while testing.

amorey:
Added in https://github.com/kubetail-org/kubetail/pull/222 with @sureshopsgeek as co-author

PR #193, opened 2025-04-13
feat: Dashboard enhancement.

victorchrollo14:
1) added a proper sidebar to switch between workloads and made it toggleable.
2) aligned the tables to adapt to the whole screen, but columns across tables are misaligned ( needs to be fixed in a upcoming pr ).
3) added a proper no results ui with some icon.
4) added search input and filter button on top of each table,only the ui, functionality in the next pr.

### DIFF
--- a/dashboard-ui/src/components/widgets/SettingsDropdown.tsx
+++ b/dashboard-ui/src/components/widgets/SettingsDropdown.tsx
@@ -36,7 +36,9 @@ const SettingsDropdown = () => {
     <>
       <DropdownMenu>
         <DropdownMenuTrigger asChild>
-          <Button intent="outline" size="sm"><SettingsIcon size={18} strokeWidth={1.5} /></Button>
+          <Button intent="outline" size="sm" className="py-[10px]">
+            <SettingsIcon size={18} strokeWidth={1.5} />
+          </Button>
         </DropdownMenuTrigger>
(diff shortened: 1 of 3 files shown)


victorchrollo14:
> * "Kubetail" in "Kubetail Dashboard" isn't necessary because the logo in the top-left says "Kubetail"
> * Remove console logging debug messages (e.g. "toggle sidebar")
> * Is search supposed to work? When I start typing nothing happens and then when I hit enter the browser navigates to the logging console
> * The filter dropdown is using placeholder values (filter, filter 1, filter 2)

will fix these up. guess it's better to remove the filter and search part in this pr and make it in the next one, as here we've just added the ui.

victorchrollo14:
> * I think we can pack the table rows a bit more densely. What do you think?
not sure, I increased the padding of each table cell, so it wouldn't look odd when we stretch it out to the whole screen. we could reduce the padding a bit more I guess.

victorchrollo14:
> * What do you think about re-using the sidebar toggle button instead of moving the toggle button to a different place? Usually when I click on a button with an undo option I expect it to be in the same place as the original.

That was what I thought initially, but couldn't figure out a good place to place the toggle button so went with this. let me know if you find any ui references, for sidebar and toggle part. I'll try searching and see if we can place it in way it doesn't look odd.

PR #207, opened 2025-04-20
feat: More improvements in the dashboard.

victorchrollo14:
- changed the sidebar toggle part.
- add a countBadge on sidebar items.
- show no resource found ui inside the table.
- hide show more option and show the whole table when viewing individual workloads.
- moved the useWorkloadCounter hook to a center location to reuse in source picker and dashboard sidebar.

https://github.com/user-attachments/assets/0998fa10-eac9-4f62-8e0d-ce3178a4873c

### DIFF
--- a/dashboard-ui/src/pages/home.tsx
+++ b/dashboard-ui/src/pages/home.tsx
@@ -41,7 +41,7 @@ import SettingsDropdown from '@/components/widgets/SettingsDropdown';
 import * as dashboardOps from '@/lib/graphql/dashboard/ops';
 import { useListQueryWithSubscription, useLogMetadata } from '@/lib/hooks';
 import { joinPaths, getBasename, cn } from '@/lib/util';
-import { Workload, allWorkloads, iconMap, labelsPMap } from '@/lib/workload';
+import { SidebarResources, Workload, iconMap, labelsPMap, sidebarResources } from '@/lib/workload';
 
 /**
  * Shared variables and helper methods
@@ -61,6 +61,10 @@ type ContextType = {
   setKubeContext: React.Dispatch<React.SetStateAction<string | undefined>>;
   workloadFilter?: Workload;
   setWorkloadFilter: React.Dispatch<React.SetStateAction<Workload | undefined>>;
+  sidebarOpen: boolean;
+  setSidebarOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  sidebar: SidebarResources,
+  setSidebar: React.Dispatch<React.SetStateAction<SidebarResources>>;
 };
 
 const Context = createContext({} as ContextType);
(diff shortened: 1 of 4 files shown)


dashboard-ui/src/pages/home.tsx
```diff
@@ -400,8 +407,11 @@ const DisplayItems = ({
     });
   }
 
+  // not showing all the rows at once when user selects all workloads filter
+  const allWorkloadView = useMemo(() => workloadFilter === undefined, [workloadFilter]);
```
amorey: Why do we need useMemo() here?

dashboard-ui/src/pages/home.tsx
```diff
@@ -400,8 +407,11 @@ const DisplayItems = ({
...
+
   // handle show some-or-all
-  const [showAll, setShowAll] = useState(false);
+  const [showAll, setShowAll] = useState(!allWorkloadView);
```
amorey, line 405: Do we need to track the state of `showAll`? Why not just use the value of `workloadFilter` itself?

dashboard-ui/src/pages/home.tsx
```diff
@@ -50,6 +50,15 @@ import { Workload, allWorkloads, iconMap, labelsPMap } from '@/lib/workload';
 const basename = getBasename();
 
 const defaultKubeContext = appConfig.environment === 'cluster' ? '' : undefined;
+const defaultSidebarItems: Map<Workload, undefined> = new Map([
```
amorey: It looks like you're using`defaultSidebarItems` to list the workloads in the sidebar while waiting for the counter data to finish loading. Instead of creating a new map to use as a placeholder, you can use `allWorkloads` to iterate through the workloads and then key into counter map when it's available. You can also take advantage of the hook's `loading` variable. For example something like this should work:

```javascript
allWorkloads.map((workload) => (
  <div>
    <div>{labelsPMap[workload]}</div>
    {loading && (<div>count: {counter.get(workload) ?? 0}</div>)}
  </div>
))
```

dashboard-ui/src/pages/home.tsx
victorchrollo14: Probably not needed, I guess. Thought we would end up with an infinite loop. Will remove this.

dashboard-ui/src/pages/home.tsx
victorchrollo14: That sounds good too, will try this

dashboard-ui/src/pages/home.tsx
victorchrollo14: Ahh, this looks so much better.

dashboard-ui/src/pages/home.tsx
```diff
@@ -817,10 +808,10 @@ const Sidebar = () => {
   const { workloadFilter, setWorkloadFilter, kubeContext, namespace } = useContext(Context);
 
   // kubeContext sometimes is undefined
-  const { counter } = useWorkloadCounter(kubeContext ?? '', namespace);
+  const { loading, error, counter } = useWorkloadCounter(kubeContext ?? '', namespace);
 
-  // the counter returns a empty map sometimes
-  const sidebarItems = counter.size === 0 ? Array.from(defaultSidebarItems) : Array.from(counter.entries());
+  // using some default sidebar values during data loading and error states
+  const sidebarItems: [Workload, number][] = loading || error ? allWorkloads.map((w) => [w, 0]) : Array.from(counter.entries());
```
amorey, line 814: Nice!

PR #212, opened 2025-04-23
feat: add oidc plugin for k8s client

SoulKyu:
Hello ! 

Just a simple fix that permit to handle OIDC connection with the kubernetes client !

Without this fix any oidc cluster will be failed with error : `Error: no Auth Provider found for name "oidc"` 

Its a very simple fix and hope you will approve it 

thanks you

### DIFF
--- a/modules/shared/k8shelpers/connection-manager.go
+++ b/modules/shared/k8shelpers/connection-manager.go
@@ -24,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"


amorey:
Thanks for the PR! I'm surprised that the kubernetes client doesn't import the necessary libraries on its own. Is there another way to add oidc support? It would be nice to integrate it more tightly so we can guarantee support in the future.

SoulKyu:
Kubernetes client-go's design decision to make auth providers optional plugins is actually deliberate, though it can be confusing at first.

We could do an import of all plugins : 
```go
    _ "k8s.io/client-go/plugin/pkg/client/auth/azure"
    _ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
    _ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
```

With using the initProvider : 

```go
func InitProviders() {
    // This function just needs to exist to make the package importable
}
```

And then : 
```go
func init() {
    k8sauth.InitProviders()
}
```

I think the reason plugin are not imported by default is to maintain the binary as small as possible.

I think keeping it as simple as possible is better for maintainability .

amorey:
Ahh thanks for the details. Sure, let's add the other providers too. Do you want to modify the PR?

amorey:
I merged your change to ship it quickly. Let's add the other providers in a separate pr.

amorey:
BTW - I'm looking for beta testers for our new search feature. Here it is in action:
https://www.kubetail.com/demo

It's live as a release candidate if you want to try it locally (helm chart v0.10.0-rc2). Let me know if you're available to give me some feedback! You can email me or find me in the Kubetail discord.

PR #243, opened 2025-05-03
Fix broken link in README (Tilt docs cluster setup)

sinancang:
I've noticed that the link https://docs.tilt.dev/choosing_clusters#microk8s is broken!
Here is the correct link: https://docs.tilt.dev/choosing_clusters.html#microk8s but we can also link to the general page:  https://docs.tilt.dev/choosing_clusters.html


amorey:
Nice catch! Thanks for the PR.

PR #244, opened 2025-05-03
feat: add search functionality to the dashboard page

victorchrollo14:
- adds search option where we can look at workloads based on search.
- adds a scroll on overflow x in the table, so we can scroll to the right if table get's wider.
- moved the filtering and searching to a seperate function.

### DIFF
--- a/dashboard-ui/src/mocks/home.ts
+++ b/dashboard-ui/src/mocks/home.ts
@@ -281,11 +281,3 @@ export const workloadItemMock: WorkloadItem[] = [
     },
   },
 ];
-
-export const ownerShipMapMock = new Map([
-  ['dc8fbace-67c0-43d5-a59d-aaff6dbe2d7a', ['60c83096-174e-4191-a705-1245b52a0e33', '5955f63b-b69b-45de-b2e1-2eb60e4cd15e', '9b72c3fd-c800-45ed-97e4-d7d965d9cf65']],
-  ['fada4cb6-0bb4-46f7-99cd-05498ef732f4', ['93edde53-0bb8-44e6-b271-0022abe42100', 'edc816e9-dea5-4133-b499-89984b9ebb14', '750545e3-2266-4891-b6eb-18bd7c5e4f09']],
-  ['db03b586-95df-48f3-aaeb-9e0de42d3926', ['3596ec70-0de7-40a9-90a8-d57f8931ae15', '603414f8-cdec-40dd-bbbe-7ada2473d77c', 'd0e68602-898a-44e3-abce-1d98ab2eee44',
-    'e7ab0fb4-ae9e-4ffe-82ec-899254e096b0']],
-  ['603414f8-cdec-40dd-bbbe-7ada2473d77c', ['fff3a147-8c0e-489d-983a-2a190b4b1999', 'ab79e4af-dccb-4b9b-9063-eea4a1c40e8c', 'dd4ad9ff-622b-4697-bbab-0b22fdb2031e']],
-]);
--- a/dashboard-ui/src/mocks/home.ts
+++ b/dashboard-ui/src/mocks/home.ts
@@ -15,7 +15,7 @@
 import type { MockedResponse } from '@apollo/client/testing';
 
 import * as dashboardOps from '@/lib/graphql/dashboard/ops';
-import { WorkloadItem } from '@/pages/home';
+import type { WorkloadItem } from '@/pages/home';
 
 const genericListResponse = (typename: string) => ({
   __typename: typename,
--- a/dashboard-ui/src/mocks/home.ts
+++ b/dashboard-ui/src/mocks/home.ts
@@ -15,6 +15,7 @@
 import type { MockedResponse } from '@apollo/client/testing';
 
 import * as dashboardOps from '@/lib/graphql/dashboard/ops';
+import { WorkloadItem } from '@/pages/home';
 
 const genericListResponse = (typename: string) => ({
   __typename: typename,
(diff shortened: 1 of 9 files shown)


dashboard-ui/src/pages/home.tsx
```diff
@@ -33,13 +33,19 @@ import type { SortBy } from '@kubetail/ui/elements/DataTable/Header';
 import Form from '@kubetail/ui/elements/Form';
 import Spinner from '@kubetail/ui/elements/Spinner';
 
+import FormControl from '@kubetail/ui/elements/FormControl';
```
amorey: Let's group this import with `Form` and `Spinner` above. In general we use four categories of imports grouped separately at the top of every page: standard library, third-party, kubetail-external, internal

dashboard-ui/src/pages/home.tsx
```diff
@@ -33,13 +33,19 @@ import type { SortBy } from '@kubetail/ui/elements/DataTable/Header';
...
+import { useDebounceCallback } from 'usehooks-ts';
```
amorey: This should be grouped with third-party imports (e.g. 'lucide-react')

dashboard-ui/src/pages/home.tsx
```diff
@@ -65,6 +71,19 @@ type ContextType = {
   setWorkloadFilter: React.Dispatch<React.SetStateAction<Workload | undefined>>;
   sidebarOpen: boolean;
   setSidebarOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  search: string;
+  setSearch: React.Dispatch<React.SetStateAction<string>>;
+};
+
+type WorkloadItem = {
```
amorey: You can re-use the graphql typescript definitions like this:
```typescript
import type {
  HomeCronJobsListItemFragmentFragment,
  HomeDaemonSetsListItemFragmentFragment,
  ...  
} from '@/lib/graphql/dashboard/__generated__/graphql';


type WorkloadItem = HomeCronJobsListItemFragmentFragment | HomeDaemonSetsListItemFragmentFragment | ...;
```

dashboard-ui/src/pages/home.tsx
```diff
@@ -198,6 +228,30 @@ function useLogFileInfo(uids: string[], ownershipMap: Map<string, string[]>) {
   return logFileInfo;
 }
 
+/**
+ * function to apply filters and search
+ */
+
+function useSearchAndFilter(fetching: boolean, items: WorkloadItem[] | null | undefined, search: string, namespace: string): undefined | WorkloadItem[] {
```
amorey: When a function name starts with "use" it gets treated as a hook by react but this particular function isn't currently using any hook functionality so it isn't necessary.

dashboard-ui/src/pages/home.tsx
```diff
@@ -274,6 +328,33 @@ const KubeContextPicker = ({
...
+    <div
+      className={cn(
+        'search-input relative',
+      )}
+    >
+      <FormControl
+        id="search-box"
+        onChange={(e) => deboucedSearch(e.target.value)}
+        type="search"
+        onKeyDown={(e) => { if (e.key === 'Enter') { e.preventDefault(); } }} // prevent form submission on enter
```
amorey: If you wrap `<FormControl>` in a `<Form onSubmit={(ev) => ev.preventDefault}>` you can avoid using `onKeyDown`.

dashboard-ui/src/pages/home.tsx
```diff
@@ -326,38 +401,33 @@ type DisplayItemsProps = {
...
+  const { kubeContext, search, workloadFilter } = useContext(Context);
 
-    // remove items not in filtered namespace
-    return (namespace === '' || item.metadata.namespace === namespace);
-  });
+  // destructuring items array due to read-only array issue
+  const filteredItems = items ? [...items] : [];
 
-  const ids = filteredItems?.map((item) => item.metadata.uid) || [];
+  const ids = items?.map((item) => item.metadata.uid) || [];
```
amorey: Why do you need the list of unfiltered item ids instead of filtered item ids here?

dashboard-ui/src/pages/home.test.tsx
```diff
@@ -32,3 +42,91 @@ describe('home page', () => {
...
+  it('returns an empty array if parentId is not present in the ownership map', () => {
+    const result = getContainerIDs('dc8fbace-67c0-aaff6dbe2d7a-id-not-present', ownershipMapMock);
+
+    expect(result).toEqual([]);
+  });
+
+  it('returns all the values of a parentId when none of the childIds have grandChildIds', () => {
+    const parentId = 'db03b586-95df-48f3-aaeb-9e0de42d3926';
+    const result = getContainerIDs(parentId, ownershipMapMock);
+    const output = ownershipMapMock.get(parentId);
```
amorey: using the variable name `expected` is clearer

dashboard-ui/src/pages/home.test.tsx
```diff
@@ -32,3 +42,91 @@ describe('home page', () => {
...
+});
+
+describe('getContainerIDs', () => {
+  it('returns an empty array if parentId is not present in the ownership map', () => {
+    const result = getContainerIDs('dc8fbace-67c0-aaff6dbe2d7a-id-not-present', ownershipMapMock);
+
+    expect(result).toEqual([]);
+  });
+
+  it('returns all the values of a parentId when none of the childIds have grandChildIds', () => {
```
amorey: "returns all the values of a parentId" is a bit confusing. maybe something more like "returns immediate children of a parent"...

dashboard-ui/src/pages/home.test.tsx
```diff
@@ -32,3 +42,91 @@ describe('home page', () => {
...
+    const parentId = 'db03b586-95df-48f3-aaeb-9e0de42d3926';
+    const result = getContainerIDs(parentId, ownershipMapMock);
+    const output = ownershipMapMock.get(parentId);
+
+    expect(result).toEqual(output);
+  });
+
+  it('returns parentId values merged with childId values, excluding childId when the childId has grandChildIds', () => {
```
amorey: can you try to simplify this statement?

dashboard-ui/src/mocks/home.ts
```diff
@@ -15,6 +15,7 @@
 import type { MockedResponse } from '@apollo/client/testing';
 
 import * as dashboardOps from '@/lib/graphql/dashboard/ops';
+import { WorkloadItem } from '@/pages/home';
```
amorey: since this is a typescript type and not a javascript object you should use
```
import type { WorkloadItem } from '@pages/home';
```

From ChatGPT:
1. Tells TS “this is a type‐only import” so if you’ve got a build step (e.g. Babel) that doesn’t automatically erase unused imports, you won’t end up with a dead import at runtime.
2. Makes your intent explicit (“I’m not pulling in any value here”), which can help tools and future readers.

dashboard-ui/src/pages/home.tsx
```diff
@@ -83,9 +99,18 @@ function getContainerIDs(
...
   });
+
   return containerIDs;
 }
 
+/**
+ * Checks if all provided arrays are either undefined or empty
+ */
+
+export function noSearchResults(...arrays: (WorkloadItem[] | undefined)[]) {
```
amorey: React hot module replacement isn't working currently because apparently it only works if all exported objects are react components. To fix this we should move the helpers to a library method (e.g. `/src/lib/home.ts`) then that should fix the problem.

Issue #245, opened 2025-05-03
Offline clusters mess up with running cluster


runningman84:
I have multiple context in my kubeconfig. The first context is a running cluster the other two contexts are test clusters which are offline most of the time. If I start kubetail I can see the pods of the first cluster, if I click on the logs of any pod I see a lot of errors in the log and I do not see anything because it seems that kubetail is trying to access the offline clusters:

```
➜  ~ kubetail serve
2025-05-03T17:08:02.729904665+02:00 INF Loaded default kubeconfig: /home/phil/.kube/config
2025-05-03T17:08:02.731471525+02:00 INF Started kubetail dashboard on http://localhost:7500/
W0503 17:08:05.779496    6619 reflector.go:569] pkg/mod/k8s.io/client-go@v0.32.3/tools/cache/reflector.go:251: failed to list *v1.EndpointSlice: Get "https://10.0.30.61:6443/apis/discovery.k8s.io/v1/namespaces/kubetail-system/endpointslices?labelSelector=kubernetes.io%2Fservice-name%3Dkubetail-cluster-api&limit=500&resourceVersion=0": dial tcp 10.0.30.61:6443: connect: no route to host
E0503 17:08:05.779546    6619 reflector.go:166] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.32.3/tools/cache/reflector.go:251: Failed to watch *v1.EndpointSlice: failed to list *v1.EndpointSlice: Get \"https://10.0.30.61:6443/apis/discovery.k8s.io/v1/namespaces/kubetail-system/endpointslices?labelSelector=kubernetes.io%2Fservice-name%3Dkubetail-cluster-api&limit=500&resourceVersion=0\": dial tcp 10.0.30.61:6443: connect: no route to host" logger="UnhandledError"
W0503 17:08:09.166117    6619 reflector.go:569] pkg/mod/k8s.io/client-go@v0.32.3/tools/cache/reflector.go:251: failed to list *v1.EndpointSlice: Get "https://10.0.30.61:6443/apis/discovery.k8s.io/v1/namespaces/kubetail-system/endpointslices?labelSelector=kubernetes.io%2Fservice-name%3Dkubetail-cluster-api&limit=500&resourceVersion=0": dial tcp 10.0.30.61:6443: connect: no route to host
[...]

PR #246, opened 2025-05-03
fix: supress kubernetes client-go background errors in cli and fix mutex race condition in health checker

amorey:
This PR fixes two issues:
1) It supresses kubernetes client-go background errors in the CLI (e.g. informer cache errors)
2) It prevents a condition where health check workers for different kube contexts were competing for the same mutex lock causing unhealthy clusters to hold up health checks for other clusters

Note: This is was done quickly so we should revisit both changes later. In particular, we need to figure out a way to expose kubernetes client-go background errors without overwhelming stdout.

Fixes https://github.com/kubetail-org/kubetail/issues/245

### DIFF
--- a/modules/cli/cmd/serve.go
+++ b/modules/cli/cmd/serve.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


amorey:
@rxinui FYI

PR #254, opened 2025-05-07
[Low Hanging Fruit] rgkl - Run Rust formatter and linter (clippy), incorporate changes

sinancang:
Mostly stylistic changes to adhere to Rust coding standards outlined by `cargo fmt` and `cargo clippy`. I thought most of the clippy suggestions were good, except for its complaint on using all caps on CRI (I think that is a pedantic warning and not really applicable here where CRI is a standard abbreviation), so I've explicitly suppressed that warning.

The only functional change is un-commenting `term_tx.send()` in `main.rs`, as the early-termination logic of `TermReader` relies on this. @amorey please do to let me know if there was a specific reason why that part was commented out.

Thanks!

### DIFF
--- a/crates/rgkl/src/main.rs
+++ b/crates/rgkl/src/main.rs
@@ -110,10 +110,10 @@ fn main() -> ExitCode {
             let mut stdout = stdout().lock();
             match stream_forward::run(
                 file,
-                start_time.clone(),
-                stop_time.clone(),
+                *start_time,
+                *stop_time,
                 grep,
-                follow_from.clone(),
+                *follow_from,
                 term_rx,
                 &mut stdout,
             ) {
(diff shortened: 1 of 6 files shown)


crates/rgkl/src/main.rs
```diff
@@ -84,9 +84,8 @@ fn main() -> ExitCode {
     match Signals::new([SIGINT, SIGTERM]) {
         Ok(mut signals) => {
             thread::spawn(move || {
-                for _sig in signals.forever() {
-                    //let _ = term_tx.send(());
-                    break;
+                if signals.forever().next().is_some() {
+                    let _ = term_tx.send(());
```
amorey: term_tx.send() shouldn't get called because the actual termination action is the drop(term_tx) that comes after

sinancang:
PR came to about 7 commits, probably best if we squash before merge

PR #256, opened 2025-05-07
rgkl - reduce dev env Rust compile-time by moving build step to local

sinancang:
## Changes
1. Define rgkl build step as a local resource in the Tiltfile
2. Build in debug mode instead of release mode

#### Build-time (`cluster-agent` )
- ~60s on first build
- ~5s on subsequent builds (pulls cached binary)

#### Update time
- ~60s on full re-compile
- ~5s for source files that were cached (ie reverting a change to a previous state of the source file)

#### Build-time
- `build-rgkl`
    - ~30s on first build
    - <1s on subsequent builds (pulls cached binary)
- `cluster-agent` (runs after `build-rgkl` )
    - ~4s always (Go build)

#### Update time
- `build-rgkl`
    - ~30s on full initial build
    - ~3s on code updates
- `cluster-agent` (runs after `build-rgkl` )
    - ~4s regardless

## Concerns
This does bring in a few more dependencies into setting up the dev environment. Namely, the user now needs to install a **protobuf compiler** and **cross-compilation tools targeting amd64** to their local machine. My apprehension is that, for someone not interested in developing the Rust sub-component, these dependencies will be irrelevant bloat to their system and extra headache that they don't really care about.

To address this, we can potentially set this up as an opt-in configuration of the dev-env, for people specifically developing the Rust code. Would be interested in approaches on how to achieve this.

## Further Improvements
Adding support for **sccache** can increase iteration speed even further, however I haven't been able to get it running locally (I think they might not support my WSL). I intend to re-visit this down the line as part of potentially another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,13 +183,15 @@ It also contains the source code for the Kubetail Dashboard's frontend and the R
 
 ### Setting up the Development Environment
 
-Dependencies:
+#### Dependencies:
+
 * [Tilt](https://tilt.dev/)
 * [Go](https://go.dev/)
 * [pnpm](https://pnpm.io/)
 * [ctlptl](https://github.com/tilt-dev/ctlptl) (optional)
 
-1. Create a Kubernetes dev cluster:
+
+1. Create a Kubernetes Dev Cluster
 
 ```console
 ctlptl apply -f hack/ctlptl/minikube.yaml
@@ -218,7 +220,55 @@ pnpm install
 pnpm dev
 ```
 
-Now access the dashboard at [http://localhost:5173](http://localhost:5173). 
+Now access the dashboard at [http://localhost:5173](http://localhost:5173).
+
+
+---
+
+
+### (Optional) Fast Iteration for Rust Development
+
+If you want to iterate quickly on Rust components, you can build them locally instead of through the dev containers.
+
+**1. Install Rust and Required Targets**
+
+**macOS:**
+
+```sh
+brew install protobuf
+rustup install stable
+rustup target add x86_64-unknown-linux-musl aarch64-unknown-linux-musl
+brew install FiloSottile/musl-cross/musl-cross
+```
+
+Then add the following to `~/.cargo/config.toml`:
+
+```toml
+[target.x86_64-unknown-linux-musl]
+linker = "x86_64-linux-musl-gcc"
+
+[target.aarch64-unknown-linux-musl]
+linker = "aarch64-linux-musl-gcc"
+```
+
+This ensures that `cargo` uses the correct linker when building for those targets.
+
+---
+
+**Linux:**
+
+```sh
+sudo apt-get update && sudo apt-get install -y protobuf-compiler musl-tools
+rustup install stable
+rustup target add x86_64-unknown-linux-musl aarch64-unknown-linux-musl
+```
+
+**2. Enable Local Rust Builds with Tilt**
+
+```sh
+BUILD_RUST_LOCALLY=true tilt up
+```
+
 
 ## Build
 
--- a/Tiltfile
+++ b/Tiltfile
@@ -52,23 +52,87 @@ local_resource(
   ]
 )
 
-docker_build_with_restart(
-  'kubetail-cluster-agent',
-  dockerfile='hack/tilt/Dockerfile.kubetail-cluster-agent',
-  context='.',
-  entrypoint="/cluster-agent/cluster-agent -c /etc/kubetail/config.yaml",
-  only=[
-    './crates/rgkl',
-    './proto',
-    './.tilt/cluster-agent'
-  ],
-  ignore=[
-    './crates/rgkl/target'
-  ],
-  live_update=[
-    sync('./.tilt/cluster-agent', '/cluster-agent/cluster-agent'),
-  ]
-)
+build_rust_locally = os.getenv("BUILD_RUST_LOCALLY", default='false').lower() == 'true'
+if build_rust_locally:
+  local_resource(
+    "kubetail-rgkl",
+    '''
+    set -eu
+
+    cd crates/rgkl
+
+    # ---------- Determine Target Architecture ----------
+    arch=$(uname -m)
+    case "$arch" in
+      x86_64|amd64) target_arch="x86_64" ;;
+      arm64|aarch64) target_arch="aarch64" ;;
+      *) echo "Unsupported arch: $arch"; exit 1 ;;
+    esac
+    target="${target_arch}-unknown-linux-musl"
+    export CARGO_TARGET_DIR="target/.tilt"
+
+    # ---------- Ensure musl Target is Installed ----------
+    if ! rustup target list | grep -q "${target} (installed)"; then
+      echo "📦 Installing Rust target ${target}"
+      rustup target add "${target}"
+    fi
+
+    # ---------- Build Static Binary ----------
+    echo "🔨 cargo build --target=${target}"
+    cargo build --target "${target}"
+
+    bin_path="${CARGO_TARGET_DIR}/${target}/release/rgkl"
+    out_dir="../../.tilt"
+
+    # ---------- Package Static Binary ----------
+    mkdir -p "$out_dir"
+    install -Dm755 "$bin_path" "$out_dir"
+    echo "✅ Static binary packaged in $out_dir"
+    ''',
+    deps=[
+      "./crates/rgkl/src",
+      "./crates/rgkl/Cargo.toml",
+      "./crates/rgkl/Cargo.lock",
+      "./proto",
+    ],
+  )
+
+  docker_build_with_restart(
+    'kubetail-cluster-agent',
+    dockerfile='hack/tilt/Dockerfile.kubetail-cluster-agent-local',
+    context='.',
+    entrypoint="/cluster-agent/cluster-agent -c /etc/kubetail/config.yaml",
+    only=[
+      './proto',
+      './.tilt/cluster-agent',
+      './.tilt/rgkl',
+    ],
+    live_update=[
+      sync('./.tilt/cluster-agent', '/cluster-agent/cluster-agent'),
+      sync(
+        './.tilt/rgkl',
+        '/usr/local/bin/rgkl',
+      ),
+    ]
+  )
+else:
+  docker_build_with_restart(
+    'kubetail-cluster-agent',
+    dockerfile='hack/tilt/Dockerfile.kubetail-cluster-agent',
+    context='.',
+    entrypoint="/cluster-agent/cluster-agent -c /etc/kubetail/config.yaml",
+    only=[
+      './crates/rgkl',
+      './proto',
+      './.tilt/cluster-agent'
+    ],
+    ignore=[
+      './crates/rgkl/target'
+    ],
+    live_update=[
+      sync('./.tilt/cluster-agent', '/cluster-agent/cluster-agent'),
+    ]
+  )
 
 # kubetail-dashboard
 
--- a/crates/rgkl/.cargo/config.toml
+++ b/crates/rgkl/.cargo/config.toml
@@ -1,0 +1,4 @@
+[profile.dev]
+incremental = true
+debug = true
+opt-level = 1
--- a/hack/tilt/Dockerfile.kubetail-cluster-agent
+++ b/hack/tilt/Dockerfile.kubetail-cluster-agent
@@ -2,8 +2,7 @@ FROM rust:1.84.1 AS rustbuilder
 
 WORKDIR /work
 
-RUN rustup target add x86_64-unknown-linux-musl aarch64-unknown-linux-musl
-RUN apt-get update && apt-get install -yq musl-tools protobuf-compiler
+RUN apt-get update && apt-get install -yq protobuf-compiler
 
 # Install Rust dependencies (for cache)
 COPY crates/rgkl/Cargo.toml ./crates/rgkl/Cargo.toml
@@ -17,22 +16,27 @@ COPY . .
 # Detect architecture and build statically
 RUN ARCH="$(uname -m)" && \
     if [ "$ARCH" = "x86_64" ]; then \
-      TARGET="x86_64-unknown-linux-musl"; \
+      TARGET="x86_64-unknown-linux-gnu"; \
     elif [ "$ARCH" = "aarch64" ]; then \
-      TARGET="aarch64-unknown-linux-musl"; \
+      TARGET="aarch64-unknown-linux-gnu"; \
     else \
       echo "Unsupported architecture: $ARCH"; exit 1; \
     fi && \
     cd crates/rgkl && \
+    RUSTFLAGS="-C target-feature=-crt-static" \
     cargo build --target=$TARGET && \
     mkdir -p /out && \
     cp target/$TARGET/debug/rgkl /out/rgkl
 
 # -----------------------------------------------------------
 
-FROM alpine:3.20.0
+FROM debian:bookworm-slim AS final
 
 WORKDIR /cluster-agent
+
+# Install required runtime dependencies for glibc binaries (if not already present)
+RUN apt-get update && apt-get install -y --no-install-recommends libc6 && \
+    rm -rf /var/lib/apt/lists/*
 
 COPY --from=rustbuilder /out/rgkl /usr/local/bin/rgkl
 COPY .tilt/cluster-agent .
--- a/hack/tilt/Dockerfile.kubetail-cluster-agent
+++ b/hack/tilt/Dockerfile.kubetail-cluster-agent
@@ -2,8 +2,8 @@ FROM rust:1.84.1 AS rustbuilder
 
 WORKDIR /work
 
-# System dependencies
-RUN apt-get update && apt-get install -yq patchelf protobuf-compiler
+RUN rustup target add x86_64-unknown-linux-musl aarch64-unknown-linux-musl
+RUN apt-get update && apt-get install -yq musl-tools protobuf-compiler
 
 # Install Rust dependencies (for cache)
 COPY crates/rgkl/Cargo.toml ./crates/rgkl/Cargo.toml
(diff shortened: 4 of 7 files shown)


crates/rgkl/.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"
+
+[profile.release]
+incremental = true
+debug = true
```
amorey, line 3: Doesn't setting `debug = true` here change the setting for all "release" profiles? We also use the "release" profile when we build for production: https://github.com/kubetail-org/kubetail/blob/main/build/package/Dockerfile.cluster-agent#L32

Instead, can we specifically tell cargo to use the debug profile in development? Would something like this work?
```console
 cargo build --target aarch64-unknown-linux-gnu --debug
 ```

File: Tiltfile [hunk 2]
```diff
@@ -52,21 +52,47 @@ local_resource(
...
+  mkdir -p ../../.tilt/rgkl/target
+
+  # write all build artifacts into .tilt/rgkl/target
+  export CARGO_TARGET_DIR=../../.tilt/rgkl/target
+
+  # point Cargo at the cross-C compiler
+  export CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc
+
+  # build once (no more rm -rf)
+  cargo build --target aarch64-unknown-linux-gnu
```
amorey: This assumes we always want to target aarch64 but on amd64 systems, the preferred container arch will be amd64. Can we check the host arch and choose the target accordingly?

Also, instead of building into the `.tilt/rgkl/target` directory how about just copying the final binary to `.tilt/rgkl`? That way the `docker_build_with_restart()` code will be target architecture agnostic.

This snippet from `install.sh` script could be useful:
```bash
# Detect OS and architecture

OS=$(uname -s | tr '[:upper:]' '[:lower:]')
ARCH=$(uname -m)

case "$OS" in
  mingw*|cygwin*) OS="windows" ;;
  linux) OS="linux" ;;
  darwin) OS="darwin" ;;
esac

case "$ARCH" in
  x86_64) ARCH="amd64" ;;
  arm64) ARCH="arm64" ;;
  aarch64) ARCH="arm64" ;;
esac
```

hack/tilt/Dockerfile.kubetail-cluster-agent
```diff
@@ -1,52 +1,13 @@
...
-
-COPY --from=rustbuilder /out /rgkl
+# your Go/JS cluster-agent
 COPY .tilt/cluster-agent .
 
-ENTRYPOINT ["./cluster-agent"]
-CMD []
+# pre-built Rust binary
+COPY .tilt/rgkl/target/aarch64-unknown-linux-gnu/debug/rgkl /usr/local/bin/rgkl
+RUN chmod +x /usr/local/bin/rgkl
```
amorey: Isn't the `rgkl` binary already executable?

amorey:
This is cool as a proof of concept! Since most devs probably won't need to tweak the rust code we could fence this in behind an env flag. That way most devs won't need to set up the rust toolchain.

Some issues:
* To install the macos cross-compiler I used this brew command:
    ```console
    brew tap messense/homebrew-macos-cross-toolchains git@github.com:github.com/messense/homebrew-macos-cross-toolchains.git
    brew install aarch64-unknown-linux-gnu
    ```
    Without the ssh url homebrew uses https and asks for github credentials
* The compiler builds the binary without error but when I try to use the binary to stream logs it throws this error: `read |0: file already closed`

crates/rgkl/.cargo/config.toml
sinancang: Good catch! This was an oversight. We want `profile.dev` here. Side note: can also configure optimization level explicitly via `opt-level`. As for the cargo command, simply omitting the `--release` flag will be enough to do a debug build.

Tiltfile
sinancang: Copying the final binary sounds good. Cleans up the mental clutter of the path a bit as well.
As for the architecture, I forgot that Docker containers use the same architecture as the host system. Which is kind of odd, now that I think about it, because I used a Windows 11 WSL (x86_64) during development but didn't encounter any issues at compile-time. I'm guessing the issues actually pop-up when running the binary. In any case, this should be do-able, I'll look into implementing something similar. Could simplify the dev dependencies as well.

hack/tilt/Dockerfile.kubetail-cluster-agent
sinancang: Not sure, will check!

sinancang:
Thanks! Yeah an env flag seems like a good way to go about not requiring everyone to get the deps.

As for the issues, I'll look into reproducing them. Did you try to use the binary directly? Or via the search functionality?

amorey:
Just via search. One clue is that during the compilation step I saw a note that said the linker wasn’t used. Didn’t dig further though so maybe it’s unrelated.

hack/tilt/Dockerfile.kubetail-cluster-agent
sinancang: Verified that it is indeed executable!

sinancang:
> Just via search. One clue is that during the compilation step I saw a note that said the linker wasn’t used. Didn’t dig further though so maybe it’s unrelated.

Turns out this was an unrelated issue, resolved in [this PR](https://github.com/kubetail-org/kubetail/pull/264)

sinancang:
Incorporated the change to auto-detect host arch and craft the Rust build command accordingly.

Also converted the Rust binary to be a static executable using musl. This makes rgkl self-contained, meaning that we no longer need to deal with glibc compatibility issues.

This required some changes in the cluster-agent glue code that ran the rgkl commands, which now doesn't require a loader. I imagine if we do go forward with a static executable approach, the production environment build steps will most likely also need to change. @amorey let's discuss tradeoffs associated with this direction and find out if it is right for kubetail.

As a final step (before wrapping up with documentation), I intend to look into setting up a conditional based on an ENV flag, as discussed. After that, I can take a look at testing the branch on a macOS and see exactly which dependencies we need to bring in. After identifying that, I'll update the README with steps to enable local-build for rgkl, as well as dependencies that are required if that is enabled.

crates/rgkl/src/test.txt
sinancang, line 1: Added accidentally, will remove!

modules/cluster-agent/internal/services/logrecords/logrecords.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/kubetail-org/kubetail/modules/cluster-agent/internal/helpers"
 )
 
+const rgklPath = "/usr/local/bin/rgkl"
```
sinancang: Lmk if there's a better place to store this hard-coded path

sinancang: I think we can also just run `rkgl` directly, but I haven't tested this.

amorey: Hardcoding it here is fine

amorey:
Regarding musl, I considered using musl previously but I read that it isn't as performant so I abandoned the effort. Also I think it was a pain to get it working on macOS if I remember corrently. Do you know what the performance tradeoffs are?

README.md
```diff
@@ -187,6 +187,37 @@ Dependencies:
...
+
+  * **Linux**:
+
+    ```bash
+    sudo apt-get update
+    sudo apt-get install -y gcc-aarch64-linux-gnu binutils-aarch64-linux-gnu
+    ```
+  * **macOS** (using Homebrew + osxcross):
+
+    ```bash
```
amorey:
```console
brew tap messense/homebrew-macos-cross-toolchains
brew install aarch64-unknown-linux-gnu
```

Tiltfile
```diff
@@ -52,21 +52,55 @@ local_resource(
...
+
+  # ---------------- arch → musl target ----------------
+  arch=$(uname -m)
+  case "$arch" in
+    x86_64|amd64) target_arch="x86_64"  ;;
+    aarch64|arm64) target_arch="aarch64" ;;
+    *) echo "Unsupported arch: $arch" >&2; exit 1 ;;
+  esac
+  target="${target_arch}-unknown-linux-musl"
+  export CARGO_TARGET_DIR="../../.tilt/rgkl/target"
```
amorey: Did you consider leaving the target directory as-is (cargo/rgkl/target) and then just copying the binary into .tilt after?

hack/tilt/Dockerfile.kubetail-cluster-agent
```diff
@@ -1,52 +1,12 @@
...
-    cp -v $LINKER_PATH /out/lib/ld-linux.so
-
-RUN patchelf --set-rpath "\$ORIGIN/lib" /out/rgkl
-
-# -----------------------------------------------------------
-
 FROM alpine:3.20.0
 
 WORKDIR /cluster-agent
+RUN apk add --no-cache libc6-compat
```
amorey: Do we still need this if we're using the static binary?

sinancang: Good catch, we won't if we decide to do static binaries.

Tiltfile
sinancang: Yeah that sounds doable. I'm open to investigating but to address fully it'd be helpful for me to know what the concern is with this approach?

amorey: Mainly it's just how we're handling the go build so it would be nice to keep things the same. But there might be one technical issue - tilt watches the entire directory to trigger restarts so if the rust build step writes to the directory multiple times then each time could trigger a restart (depending on how title handles it).

amorey:
Keeping the PR as simple as possible sounds good.

Do you know if musl affects regex performance? Also, the ripgrep source code has a special section that replaces the default memory allocator when you're using musl so the original author ran into performance issues though I don't know if he solved them all.

Tiltfile
sinancang: OK cool, I'll look for a way to keep the directory structures unified while also not triggerring unnecessary restarts.

sinancang:
`musl` won't affect the performance of the regex engine as it is [purely written in and optimized for Rust](https://docs.rs/regex/latest/regex/) and doesn't rely on the standard C `regex.h` header. That being said, it will affect functionality that relies on syscalls (e.g. file read/write, threading, memory allocation, etc.) which is why the author of the library might've wanted to handle memory management in a customized way.

All in all, whether something is a "performance issue" depends entirely on the use case. In our case, the cluster agent is completely I/O bound, meaning that the real bottleneck is the time spent writing to stdout from the Rust binary and reading from stdout from the Go code, as well as the time it takes for the Go code to serialize the data via gRPC and communicate it to the application. This means that any optimizations made at making the Rust code run faster is vastly overshadowed by the sluggish time it takes to write the data and read it on the other end. THIS is the performance bottleneck we need to focus on if we want to improve user experience.

BUT I'm not a big fan of adding friction to the macOS developer experience, which is why I'll investigate if cross-compiling by keeping the builds dynamic is possible and pain-free (spoiler: it should be, just a matter of getting it to work).

sinancang:
That is a good point in that there may be cases where the log being grepped through is vastly larger than the returned subset, but even then: I doubt processing speed would be more than I/O cost. Then again, I haven't ran the numbers, so I can't say for sure. Nonetheless, ideally we keep the builds dynamic so we don't have to worry about whether the cut in rgkl performance affects user experience.

amorey:
Sounds good. Long term, I'm not opposed to using musl. It would definitely simplify the docker setup. I just want to proceed with caution because that's what the ripgrep author has done too.

Tiltfile
sinancang: Done! I set the cargo target to `crates/rgkl/target` and simply copied the binary to `.tilt/rgkl` after the build is done.

sinancang:
Switched back to dynamic builds, moved cluster-agent Docker env to debian::bookworm for out-of-the-box glibc support, helped simplify the Go glue a bit (call `rgkl` directly, no need to load explicitly). Also keep the `CARGO_TARGET_DIR` the same and only copy the binary to `.tilt`.

Still a few things to address before merging:
- Use either local build or Docker build based on an ENV flag
- Update README.md

Tiltfile
```diff
@@ -52,21 +52,55 @@ local_resource(
...
-    './.tilt/cluster-agent'
-  ],
-  ignore=[
-    './crates/rgkl/target'
+    './.tilt/cluster-agent',
+    './.tilt/rgkl',
   ],
   live_update=[
-    sync('./.tilt/cluster-agent', '/cluster-agent/cluster-agent'),
+    sync('./.tilt/cluster-agent/cluster-agent', '/cluster-agent/cluster-agent'),
```
amorey: Did the location of the cluster-agent binary change? Shouldn't it be `./tilt/cluster-agent`?

sinancang: Must've edited that accidentally, mb!

sinancang: Will revert in a bit

Tiltfile
```diff
@@ -52,21 +52,55 @@ local_resource(
...
+  cd crates/rgkl
+
+  # ---------------- arch → musl target ----------------
+  arch=$(uname -m)
+  case "$arch" in
+    x86_64|amd64) target_arch="x86_64"  ;;
+    aarch64|arm64) target_arch="aarch64" ;;
+    *) echo "Unsupported arch: $arch" >&2; exit 1 ;;
+  esac
+  target="${target_arch}-unknown-linux-gnu"
```
amorey: `./target` is the default build directory so building there will conflict with manual builds. Can you do something like this?

```bash
 arch=$(uname -m)
  case "$arch" in
    x86_64|amd64) target_arch = x86_64"  ;;
    aarch64|arm64) target_arch="aarch64" ;;
    *) echo "Unsupported arch: $arch" >&2; exit 1 ;;
  esac
target="${target_arch}-unknown-linux-gnu"
cargo build --target "${target}"
cp target/${target}/debug/rgkl ../../.tilt/rgkl
```

amorey:
> Switched back to dynamic builds, moved cluster-agent Docker env to debian::bookworm for out-of-the-box glibc support, helped simplify the Go glue a bit (call `rgkl` directly, no need to load explicitly). Also keep the `CARGO_TARGET_DIR` the same and only copy the binary to `.tilt`.
> 
> Still a few things to address before merging:
> 
>     * Use either local build or Docker build based on an ENV flag
> 
>     * Update README.md

The docker image we use in production is a "scratch" image that requires a linker to run rgkl (see [here](https://github.com/kubetail-org/kubetail/blob/main/build/package/Dockerfile.cluster-agent#L77)) which is why the docker setup in development looks like it does. If we change how we call the binary in development then we either have to change the docker image in production or set up a development/production flag for the code. I think the easiest thing to do for now is to keep the rgkl docker execution environment the same and wait until we make a longer-term decision about the cluster agent (rust, musl, etc.).

Tiltfile
sinancang: `export CARGO_TARGET_DIR="crates/rgkl/target/.tilt"` is probably what you were looking for here :)

sinancang:
Interesting. There is a way to copy the linker and it's dependencies, but it is not pretty:
```
  # ---------------- collect runtime -------------------
  bin_path="${CARGO_TARGET_DIR}/${target}/debug/rgkl"
  out_dir="../../.tilt/rgkl"
  lib_dir="${out_dir}/lib"
  mkdir -p "${lib_dir}"

  # 1. find the exact loader built against
  interp_path=$(ldd "${bin_path}" | awk "/${interp_pattern}/{print \\$1}")
  cp "${interp_path}" "${lib_dir}/"

  # 2. copy every direct shared-lib dependency
  ldd "${bin_path}" | awk '/=>/{print $(NF-1)}' | while read so; do
      cp -L "$so" "${lib_dir}/"
  done

  # 3. patch the ELF header so the kernel uses our private loader + RPATH
  patchelf --set-interpreter "\\$ORIGIN/lib/$(basename "${interp_path}")" \\
           --set-rpath "\\$ORIGIN/lib" \\
           "${bin_path}"
```

The alternative is using a container image with glibc on it in production as well. I imagine the SCRATCH image uses a lot less memory so is more efficient, but are there any other limiting factors that requires the container to be a scratch image?

sinancang:
I did get it to work reliably on Linux, but since macOS doesn't have `ldd` or `patchelf`, we can't rely on this hacky method to support development in both Linux and macOS.

So, what are the alternatives?
1. Keep dynamic loading, use a docker image that has a later version of glibc and a linker in both development and production. Con: production container is slightly larger. Pro: we don't need to worry no more about funny bundling/packaging steps.
2. Migrate to static builds with musl. Pro: much simpler builds, self-contained, light binary; less headaches. Con: benchmarks using musl vs glibc are inconclusive. Slight cut in rgkl performance is expected. Will mostly be overshadowed by I/O cost of Go cluster-agent read and gRPC send.
3. Keep dev-env builds completely container based. Pro: keep build-system as-is. Con: slower iteration in dev-env, worse developer experience.

sinancang:
Verified on Linux both prod and dev builds (local and docker-based). All works well.
However, having some issues on macOS cross-compiling for linux-musl. Investigating further.

sinancang:
The local-build instructions seem to be accurate for macOS as well.

BUT, apparently the macOS linker and cross-compilation dependencies are required for Docker-based builds as well. This isn't ideal. The goal is to have the Docker build be fully no-dependency required (or minimal). I'm going to investigate using Cross at this stage to see if it provides a minimal friction path to containerized builds.

sinancang:
Decided to use dynamic linking (like in prod) for dev-env containerized workflows.

So far, I've verified build and functionality of the following:
- dev-env container-based builds function as intended, on macOS and Linux. Doesn't require any external dependencies.
- dev-env local builds work too, on macOS and Linux. Does require some setup, steps outlined in README.md does help.
- prod builds and works, on Linux (slightly modified the dev Tiltfile to work to prod Dockerfile, verified the build and UI behavior). Also: the build is fully containerized so it should be host OS agnostic, which is why I didn't test on macOS.

Currently the build strategy is:
- Use musl for local dev-env builds, to avoid bundling the user linker and glibc with the binary, which is do-able in Linux but presents some challenges in macOS.
- Use dynamic builds and an image with a linker and glibc in non-local dev and prod builds. We can do this because since we provision the build image, we have full control of the dependencies that go into the executable. This means we can reliably ensure that the same dependencies are satisfied in the runtime image.

I think this is all I wanted to get done as part of this PR. @amorey let me know if there are anything else you'd like addressed.  Otherwise we should be good to squash and merge. Thanks!

amorey:
Thanks! I merged this into a `rgkl-compile-time` branch in the main repo. I'll take a deeper dive and create a PR if necessary before merging into `main`.

sinancang:
Sounds good, thanks! Let me know if I can help :)

Issue #183, opened 2025-04-06
Improve source picker UI


amorey:
The source picker in the logging console could use some UI love:

<img width="1145" alt="Image" src="https://github.com/user-attachments/assets/3504f1cc-fe55-4353-aaf8-8f0f5c8d0124" />

chiefkarim:
The console looks perfect for me as a developer (not much of a designer) except for the cancel button. Maybe we flip the font white.

abdelkd:
As @chiefkarim noted, but I'd say adding a right border to the side bar and reducing the gap between the check box and names would be good. As for cancel button it would be nice to make it white background with black text and border.

amorey:
Those changes sound good to me! Would either of you like to submit a PR? The code is here:
https://github.com/kubetail-org/kubetail/blob/main/dashboard-ui/src/components/widgets/SourcePickerModal.tsx

PR #277, opened 2025-05-11
Improve source picker UI

christianvmm:
Added a fixed height to the sources list container.
Implemented a sticky header for the table.
Made the entire row clickable for selection.


amorey:
Awesome! The sticky header and the clickable rows are nice usability improvements. What do you mean by a fixed height? When I change the browser height the source picker height is also changing.

BTW, linter is complaining because of single quotes in TSX attribute values. We use the airbnb styleguide: https://github.com/airbnb/javascript

rxinui:
Fixes #183

PR #278, opened 2025-05-11
Fix workload counts in dashboard sidebar

christianvmm:


amorey:
@victorchrollo14 Can you take a look at this PR when you get a chance? @christianvmm is working on updating the workload counts in the sidebar to reflect the results of a search and since the GraphQL "counter" queries don't allow search, he's using your `applySearchAndFilter()` method. What's the best way to centralize the result of `applySearchAndFilter()` so we can do it once and share the result with the sidebar and the main panel?

victorchrollo14:
sure. but there would be one issue if we are modifying the counts on the sidebar.  if we are viewing an individual filter, the count for all the workloads would change based on the search item, is that a behaviour we would want or would it be nice to only update that particular workload count in the sidebar.

dashboard-ui/src/pages/home.tsx
```diff
@@ -866,18 +866,28 @@ const CountBadge = ({ count, workload, workloadFilter }: { count: number, worklo
...
+
+  const counts = {
+    cronjobs: applySearchAndFilter(cronjobs.fetching, cronjobs.data?.batchV1CronJobsList?.items, search, namespace),
+    daemonsets: applySearchAndFilter(daemonsets.fetching, daemonsets.data?.appsV1DaemonSetsList?.items, search, namespace),
+    pods: applySearchAndFilter(pods.fetching, pods.data?.coreV1PodsList?.items, search, namespace),
+    jobs: applySearchAndFilter(jobs.fetching, jobs.data?.batchV1JobsList?.items, search, namespace),
+    deployments: applySearchAndFilter(deployments.fetching, deployments.data?.appsV1DeploymentsList?.items, search, namespace),
+    replicasets: applySearchAndFilter(replicasets.fetching, replicasets.data?.appsV1ReplicaSetsList?.items, search, namespace),
+    statefulsets: applySearchAndFilter(statefulsets.fetching, statefulsets.data?.appsV1StatefulSetsList?.items, search, namespace),
+  }
```
victorchrollo14: hey @christianvmm. can we move this to a hook and reuse it in both the sidebar and displayworkloads component.

dashboard-ui/src/pages/home.tsx
```diff
@@ -196,6 +196,36 @@ function useLogFileInfo(uids: string[], ownershipMap: Map<string, string[]>) {
...
+    filterCronJobs,
+    filterDaemonsets,
+    filterDeployments,
+    filterJobs,
+    filterPods,
+    filterReplicasets,
+    filterStatefulsets,
+  };
+}
+
```
victorchrollo14: looks good.

dashboard-ui/src/pages/home.tsx
```diff
@@ -196,6 +196,44 @@ function useLogFileInfo(uids: string[], ownershipMap: Map<string, string[]>) {
...
```
victorchrollo14: the logic works fine. but the issue is we are still doing the filtering, when we don't need it.
you could add some conditionals inside the hook to filter only when required and return the original data, when filtering isn't required. somethink like below might work.

```
  const { kubeContext, search, namespace, workloadFilter } = useContext(Context);

const shouldFilter = (workloadType: Workload) => workloadFilter === undefined || workloadFilter === workloadType;

  const filterCronJobs = shouldFilter(Workload.CRONJOBS)
    ? applySearchAndFilter(cronjobs.fetching, cronjobs.data?.batchV1CronJobsList?.items, search, namespace)
    : cronjobs.data?.batchV1CronJobsList?.items;
```

victorchrollo14 (approved):
looks good to me @amorey. we can merge this up.

PR #279, opened 2025-05-11
new: implement kubeconfig flag as global to CLI

rxinui:
## Summary

Pre-work: add `kubeconfig` flag as persistent/global to all kubetail CLI commands.

## Help wanted

shared helm.go is using `helm` cli which makes use of `*cli.EnvSettings`. This settings is used to define `kubeContext` from `--kube-context` flag. Is that the way `helm` makes use of the context? If so, I believe this _must be_ the same logic to give helm a kubeconfig.

Could you please double check @amorey

### DIFF
--- a/modules/cli/cmd/cluster_install.go
+++ b/modules/cli/cmd/cluster_install.go
@@ -40,14 +40,14 @@ var clusterInstallCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		// Get flags
 		Kubeconfig, _ := cmd.Flags().GetString(KubeconfigFlag)
-		kubeContext, _ := cmd.Flags().GetString(KubecontextFlag)
+		kubeContext, _ := cmd.Flags().GetString(KubeContextFlag)
 		//name, _ := cmd.Flags().GetString("name")
 		//namespace, _ := cmd.Flags().GetString("namespace")
 		name := helm.DefaultReleaseName
 		namespace := helm.DefaultNamespace
 
 		// Init client
-		client := helm.NewClient(helm.WithKubecontext(kubeContext), helm.WithKubeconfig(Kubeconfig))
+		client := helm.NewClient(helm.WithKubeContext(kubeContext), helm.WithKubeconfig(Kubeconfig))
 
 		// Install
 		release, err := client.InstallLatest(namespace, name)
@@ -65,7 +65,7 @@ func init() {
 	// serveCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	flagset := clusterInstallCmd.Flags()
 	flagset.SortFlags = false
-	flagset.String(KubecontextFlag, "", "Name of the kubeconfig context to use")
+	flagset.String(KubeContextFlag, "", "Name of the kubeconfig context to use")
 	//flagset.String("name", helm.DefaultReleaseName, "Release name")
 	//flagset.StringP("namespace", "n", helm.DefaultNamespace, "Namespace to install into")
 }
--- a/modules/cli/cmd/cluster_install.go
+++ b/modules/cli/cmd/cluster_install.go
@@ -39,15 +39,15 @@ var clusterInstallCmd = &cobra.Command{
 	Long:  clusterInstallHelp,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Get flags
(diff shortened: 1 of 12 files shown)


amorey:
Will take a look asap ❤️

modules/cli/cmd/cluster_install.go
```diff
@@ -65,7 +65,7 @@ func init() {
 	// serveCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	flagset := clusterInstallCmd.Flags()
 	flagset.SortFlags = false
-	flagset.String("kube-context", "", "Name of the kubeconfig context to use")
+	flagset.String(KubecontextFlag, "", "Name of the kubeconfig context to use")
```
amorey: If `KubecontextFlag` is global now can we add it to `rootCmd` instead?

modules/shared/helm/helm.go
```diff
@@ -378,13 +378,28 @@ func (c *Client) getChart(pathOptions action.ChartPathOptions) (*chart.Chart, er
 	return chart, err
 }
 
+type ClientOption func(c *Client)
+
 // Return new client
-func NewClient(kubeContext string) (*Client, error) {
-	settings := cli.New()
+func NewClient(options ...ClientOption) *Client {
```
amorey, line 384: Should we make `kubeconfig` and `kubeContext` required arguments rather than optional arguments? If the helm client is called without a `WithKubeconfig()` option, what does it do? I'm going back and forth myself but in general idomatic Go code favors explicit rather than optional arguments and in this case I think it's good to make the caller specify which cluster the actions will be performed on.

amorey: I thought about it some more and the Option pattern makes sense here because `EnvSettings` already comes with defaults and this is just overriding them.

modules/shared/helm/helm.go
```diff
@@ -378,13 +378,28 @@ func (c *Client) getChart(pathOptions action.ChartPathOptions) (*chart.Chart, er
...
+func WithKubeconfig(kubeconfig string) ClientOption {
+	return func(c *Client) {
+		c.settings.KubeConfig = kubeconfig
 	}
 
-	return &Client{settings}, nil
+}
+
+// Option Kubecontext
+func WithKubecontext(kubeContext string) ClientOption {
```
amorey: What do you think about capitalizing "context" here to be consistent with `kubeContext` elsewhere?

amorey:
Overall looks good. Let's add a couple of simple tests just to make sure that the `Option` instances are setting the `EnvSettings` attributes we expect.

modules/shared/helm/helm.go
rxinui: Yes, the option pattern makes sense because of the EnvSettings.

In addition, I wanted to call you out on the `Client` struct.
Isn't it *too much* to create a struct which contains a ONLY `EnvSettings`

I would be idiomatic by doing this instead [here](https://github.com/kubetail-org/kubetail/blob/573267ebf5df1f4c0c8e8103cea0c93b62a9ad3d/modules/shared/helm/helm.go#L51C1-L54C2):
```go
// Client
type Client struct {
	*cli.EnvSettings
}
```

instead of being verbose with `c.settings` which is confusing to me

modules/shared/helm/helm.go
rxinui: Yes will do!

rxinui:
> Overall looks good. Let's add a couple of simple tests just to make sure that the `Option` instances are setting the `EnvSettings` attributes we expect.

Don't you trust me on the Option pattern? Hmm.. that's a shame, you know I learned from the best.

ps: *will do add test just in case* 😉

amorey:
I trust you just fine 🧐 We just need to start laying down some test cases so we can modify the code more confidently in the future.

modules/shared/helm/helm.go
amorey: When I created the struct I wasn't sure what attributes/methods it would have so I kept it as generic as possible. Wouldn't passing the client instance to helm methods like this be less intuitive? https://github.com/kubetail-org/kubetail/blob/573267ebf5df1f4c0c8e8103cea0c93b62a9ad3d/modules/shared/helm/helm.go#L228

PR #283, opened 2025-05-12
🐋 Modify cluster-agent docker files so that `rgkl` can be called without linker

amorey:
## Summary

This PR modifies the production and development cluster-agent docker files such that we can build a glibc dynamically-linked `rgkl` binary that can be run from go code without the explicit use of a linker. This simplifies the docker files themselves and also makes it easier to use statically-linked `rgkl` binaries in development.

## Changes

Modify cluster-agent production and development docker files so that required libraries to run rgkl are copied into the final image in their expected places so that `rgkl` can be run from go code without the explicit use of a linker.

## Submitter checklist

- [x] Add the correct emoji to PR title
- [x] Link the issue number to *Fixes #*
- [x] Explain the changes made
- [ ] Sign the commit using for [DCO](https://github.com/apps/dco) 
- [ ] Rebase branch to HEAD

---

Signed-off-by: Andres Morey <andresmarcel@gmail.com>


Tiltfile
```diff
@@ -52,42 +52,31 @@ local_resource(
...
     out_dir="../../.tilt"
-
-    # ---------- Package Static Binary ----------
     mkdir -p "$out_dir"
-    install -Dm755 "$bin_path" "$out_dir"
-    echo "✅ Static binary packaged in $out_dir"
+    cp "target/${target}/debug/rgkl" "$out_dir"
     ''',
     deps=[
       "./crates/rgkl/src",
```
sinancang, line 82: Could add `./crates/rgkl/.cargo/config.toml` as well

crates/rgkl/.cargo/config.toml
```diff
@@ -2,3 +2,9 @@
 incremental = true
 debug = true
 opt-level = 1
+
+[target.x86_64-unknown-linux-musl]
```
sinancang: These linker configurations break Linux builds, as the `musl-tools` apt package contains `musl-gcc` as the linker. 
#### Naive solution
To resolve this in a clean way, we could use an environment variable to configure the linker, like so:
```toml
[target.x86_64-unknown-linux-musl]
linker = "${CARGO_MUSL_X86_64_LINKER}"

[target.aarch64-unknown-linux-musl]
linker = "${CARGO_MUSL_AARCH64_LINKER}"
```
And then add a step in the README.md to export them as follows:
```console
# macOS
export CARGO_MUSL_X86_64_LINKER="x86_64-linux-musl-gcc"
export CARGO_MUSL_AARCH64_LINKER="aarch64-linux-musl-gcc"

# Linux
export CARGO_MUSL_X86_64_LINKER="musl-gcc"
export CARGO_MUSL_AARCH64_LINKER="musl-gcc"
```
BUT, this **DOES NOT WORK** as tilt doesn't see the user's environment variables. An alternative is to make this export in the Tiltfile directly, as part of the local resource, but even then, what if a user uses a different linker than the ones we hard-code?

#### Best solution (imo)
I think the best approach here is to remain flexible, and ask the user to modify their local cargo config in `~/.carg/config.toml` and add a linker to the specified targets, if they're using macOS. Something like this should be good:

> If using macOS, make sure to configure the correct linkers for the musl targets in your `~/.cargo/config.toml`, as follows:
```toml
[target.x86_64-unknown-linux-musl]
linker = "x86_64-linux-musl-gcc"

[target.aarch64-unknown-linux-musl]
linker = "aarch64-linux-musl-gcc"
```

sinancang (changes requested):
Need to remove target linker configurations from repo as the user's platform might be using a different linker. Can be helpful to provide a way to configure the linker for macOS users, depending on how much hand-holding is needed.

Otherwise, lgtm!

sinancang (approved):

crates/rgkl/.cargo/config.toml
amorey: Cool, thanks. I updated the README to say macos users should modify their `~/.cargo/config.toml` file

PR #295, opened 2025-05-15
🐋 KUBECONFIG support on serve and logs command

### LINKED ISSUE #225: Add support for KUBECONFIG env var
Let's add support for KUBECONFIG env var to CLI. I think a prerequisite for this is adding kubeconfig support to the `cluster` command and making it global.

https://github.com/kubetail-org/kubetail/issues/184#issuecomment-2837583308

bavarianbidi:
initial support on the serve and logs subcommand for KUBECONFIG environment variable. As per spec it supports multiple files (defined in the KUBECONFIG environment variable)

Fixes #225 

## Summary

For the `serve` and `logs` subcommand it's it's now possible to either:
* explicitly set the path to the kube-config file via the `kube-config` flag
* or use the default kube-config file location (which is `$HOME/.kube/config`)
* or use all the config-files which are defined in the `$KUBECONFIG` environment variable

For the `helm` related subcommands it's not possible to use the multi-file support of the `$KUBECONFIG` variable due to the missing support in the `helm` go client (xref: https://github.com/helm/helm/issues/12742#issuecomment-2883905104)

## Changes

The type [ClientConfigLoadingRules](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L107-L133) of `client-go/clientcmd` (already used in the code-base) is getting introduced to get the context loading precedence (as described in the summary)

## Submitter checklist

- [x] Link the issue number to *Fixes #*
- [x] Rebase branch to HEAD
- [x] Squash changes into one signed, single commit ([suggested format](/.github/pull-request-commits.md))

### DIFF
--- a/modules/cli/cmd/root.go
+++ b/modules/cli/cmd/root.go
@@ -50,7 +50,7 @@ func init() {
 	// will be global for your application.
 
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cli.yaml)")
-	rootCmd.PersistentFlags().String(KubeconfigFlag, clientcmd.RecommendedHomeFile, "Path to kubeconfig file")
+	rootCmd.PersistentFlags().String(KubeconfigFlag, "", "Path to kubeconfig file")
 
 	// Cobra also supports local flags, which will only run
 	// when this action is called directly.
--- a/modules/shared/k8shelpers/kube-config-watcher.go
+++ b/modules/shared/k8shelpers/kube-config-watcher.go
@@ -15,9 +15,6 @@
 package k8shelpers
 
 import (
-	"os"
-	"path/filepath"
-	"strings"
 	"sync"
 
 	evbus "github.com/asaskevich/EventBus"
@@ -31,7 +28,7 @@ const HOMEPATH_TILDE = "~"
 
 // Represents KubeConfigWatcher
 type KubeConfigWatcher struct {
-	kubeconfigPath string
+	kubeconfigPath []string
 	kubeConfig     *api.Config
 	watcher        *fsnotify.Watcher
 	eventbus       evbus.Bus
@@ -39,29 +36,7 @@ type KubeConfigWatcher struct {
 }
 
 // Creates new KubeConfigWatcher instance
-func NewKubeConfigWatcher(kubeconfigPath string) (*KubeConfigWatcher, error) {
-	// Initialize kube config
-	// TODO: Handle missing kube config files more gracefully
-	var kubeConfig *api.Config
-	var err error
-
-	if kubeconfigPath == "" {
-		kubeconfigPath = clientcmd.RecommendedHomeFile
-	} else if strings.HasPrefix(kubeconfigPath, HOMEPATH_TILDE) {
-		homeDir, err := os.UserHomeDir()
-		if err != nil {
-			zlog.Error().Msg("Kubeconfig path is corrupted. Please provide a valid path")
-			return nil, err
-		}
-		kubeconfigPath = filepath.Join(homeDir, kubeconfigPath[2:])
-	}
-
-	kubeConfig, err = clientcmd.LoadFromFile(kubeconfigPath)
-
-	if err != nil {
-		zlog.Error().Msg("Kubeconfig is corrupted or missing. Please provide a valid kubeconfig.")
-		return nil, err
-	}
+func NewKubeConfigWatcher(kubeConfig *api.Config, kubeConfigPaths []string) (*KubeConfigWatcher, error) {
 
 	// Initialize watcher
 	watcher, err := fsnotify.NewWatcher()
@@ -69,14 +44,17 @@ func NewKubeConfigWatcher(kubeconfigPath string) (*KubeConfigWatcher, error) {
 		return nil, err
 	}
 
-	err = watcher.Add(kubeconfigPath)
-	if err != nil {
-		return nil, err
+	// watch all kubeconfig paths
+	for _, i := range kubeConfigPaths {
+		err = watcher.Add(i)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// Initialize
 	w := &KubeConfigWatcher{
-		kubeconfigPath: kubeconfigPath,
+		kubeconfigPath: kubeConfigPaths,
 		kubeConfig:     kubeConfig,
 		watcher:        watcher,
 		eventbus:       evbus.New(),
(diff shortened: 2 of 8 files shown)


rxinui:
This changes are not right. I think you are doing what has been done already.

modules/cli/cmd/root.go
```diff
@@ -50,7 +50,7 @@ func init() {
 	// will be global for your application.
 
 	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cli.yaml)")
-	rootCmd.PersistentFlags().String(KubeconfigFlag, clientcmd.RecommendedHomeFile, "Path to kubeconfig file")
+	rootCmd.PersistentFlags().String(KubeconfigFlag, "", "Path to kubeconfig file")
```
rxinui: Why this change?
By default, it should go to `clientcmd.RecommendedHomeFile`

rxinui (changes requested):

modules/cli/cmd/root.go
bavarianbidi: As we create a `ClientConfigLoadingRules` object ([client-go/clientcmd/loader.go](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L157-L178)) the default handling will be done there (which is required to generate a valid precedence)

bavarianbidi: The default is still `clientcmd.RecommendedHomeFile` (https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L48)

modules/shared/k8shelpers/kube-config-watcher.go
```diff
@@ -138,7 +116,7 @@ func (w *KubeConfigWatcher) start() {
 			case fsEv.Op&fsnotify.Create == fsnotify.Create:
 				// Load config
 				w.mu.Lock()
-				kubeConfig, err := clientcmd.LoadFromFile(w.kubeconfigPath)
+				kubeConfig, err := clientcmd.LoadFromFile(fsEv.Name)
```
bavarianbidi, line 117: This might cause issues as it will overwrite the global merged `*api.Config` object by a `*api.Config` with the configuration from the changed file ... need to fix this.

modules/cli/cmd/root.go
rxinui: I see. It is quite confusing. Not that what you have done is wrong, but more like the codeflow (reading flow) seems off. Here's my point of view:

When a dev looks at the CLI code, the cmd/ package defines the structure and high-level execution of each commands. Therefore, every defaults and flags should be on this level. In that sense, the use of `""` rather than `clientcmd.RecommendedHomeFile` is confusing. 

From a new contributor point of view, it could be confusing.

What I would recommend to do, is to keep `clientcmd.RecommendedHomeFile` for readability and ensure that the chain of predecence still works as intended.

What do you think?

CC @amorey

modules/shared/k8shelpers/kube-config-watcher.go
amorey: I think the simplest way to modify the watcher is to have it handle the config "merging" internally so that downstream consumers continue to see just one config object. If there were unit tests that would be an easier (sorry!). Do you want to try to do it as part of this PR or hand off that piece?

modules/cli/cmd/root.go
amorey: I understand the logic behind offloading as much as possible to `clientcmd` but I as a developer I prefer to see flag handling in the command's main execution loop. Would it interfere too much with the rest of the code to leave the default?

bavarianbidi: no, should work as much as it does atm.

but i will add some unit tests to make it more transparent

amorey:
@bavarianbidi As the original author of `KubeConfigWatcher` I feel responsible that there aren't any unit tests to ease development. Would it help if I upgraded `KubeConfigWatcher` to support multiple files in a separate PR that you can rebase on top of?

rxinui:
Hi @bavarianbidi 

We have gave a thought about the changes you made. Overall, it looks good and I think in the long run, we will go down that road. As of now, we are lacking of tests on the CLI and shared/ side. Therefore, it might be really risky to accept that many changes, especially at the low-level library (.ie Watcher).

With that being said, we would like to minimise the changes provided to limit breaking changes. Due to these circumstances, would it be possible to:

1. Handle the precedence between `--kubeconfig` and `$KUBECONFIG` at the Command level
2. Revert changes in the shared library, as much as possible

This proposal might (or might not) require not to leverage `spf13/viper` and might need us to write our own precedence algorithm, especially if `clientcmd` is not about to do it (to double check).

@bavarianbidi do you think you can take a look at that?

bavarianbidi:
> Hi @bavarianbidi
> 
> We have gave a thought about the changes you made. Overall, it looks good and I think in the long run, we will go down that road. As of now, we are lacking of tests on the CLI and shared/ side. Therefore, it might be really risky to accept that many changes, especially at the low-level library (.ie Watcher).
> 
> With that being said, we would like to minimise the changes provided to limit breaking changes. Due to these circumstances, would it be possible to:
> 
>     1. Handle the precedence between `--kubeconfig` and `$KUBECONFIG` at the Command level
> 
>     2. Revert changes in the shared library, as much as possible
> 
> 
> This proposal might (or might not) require not to leverage `spf13/viper` and might need us to write our own precedence algorithm, especially if `clientcmd` is not about to do it (to double check).
> 
> @bavarianbidi do you think you can take a look at that?

moved the precedence handling into our own shared/config package and added few tests to make it more transparent how the config handling works

bavarianbidi:
will rebase and continue with the kube-config-watcher soon

amorey:
Awesome! Thanks for the update.

amorey:
@bavarianbidi What do you think about splitting this up into two PR's - one to add support for `KUBECONFIG` with one path and a second to add support for multiple paths?

bavarianbidi:
> @bavarianbidi What do you think about splitting this up into two PR's - one to add support for `KUBECONFIG` with one path and a second to add support for multiple paths?

doable, for sure.
then we just use the first element of the `KUBECONFIG` slice

bavarianbidi:
can't reproduce the failed test on my machine:

> --- FAIL: TestLogMetadata (6.07s)
>    testing.go:1399: race detected during execution of test
> FAIL
> FAIL	github.com/kubetail-org/kubetail/modules/cluster-agent/internal/services/logmetadata	6.351s

@rxinui / @amorey do you have any idea why this is failing?

amorey:
https://github.com/kubetail-org/kubetail/pull/339

amorey:
Fix has been merged into `main`

rxinui (approved):
/lgtm

Good job @bavarianbidi thank you for taking our consideration into account

amorey (approved):

amorey:
Thanks @bavarianbidi! Sorry it turned out to be a bit of an iceberg. I merged into a temporary feature branch so I can fix that flakey test finally. Will merge into main asap.

PR #308, opened 2025-05-19
🐋  Linux packages build

### LINKED ISSUE #270: Add .deb and .rpm packages to release artifacts
We should add .deb and .rpm packages to the release artifacts that we attach to every new release of the Kubetail CLI. The place to do it is in the "release-cli" workflow action:
https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/release-cli.yml

victorchrollo14:
Fixes #270 

## Summary
- this pr adds 2 jobs in release ci package-rpm and package-deb.
- essentially we take the kubetail binary and package it to .deb and .rpm, so users could install the binary with the commands like ``` sudo dnf install kubetail.rpm``` or ``` sudo dpkg -i kubetail.deb```

## Changes
- adds 2 ci jobs and a spec file for rpm package.

## Submitter checklist

- [x] Rebase branch to HEAD

Hey @rxinui  and @amorey. can you give me review on this one. essentially I'm a noob at writing ci.


.github/workflows/release-cli.yml
```diff
@@ -6,14 +6,21 @@ permissions:
...
-      - 'cli/v*'
+      - "cli/v*"
 
 jobs:
   build:
     name: Build binaries
     strategy:
       matrix:
-        os: [ubuntu-24.04, ubuntu-24.04-arm, macos-latest, macos-13, windows-latest]
+        os:
```
amorey, line 16: Using "-" for vertical lists in yaml is more common:
```
os:
  - ubuntu-24.04
  - ubuntu-24.04-arm
...
```

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+        run: |
+          mkdir -p .debpkg/usr/bin
+          cp -p bin/* .debpkg/usr/bin
+
+      - name: create a .deb package
+        uses: jiro4989/build-deb-action@v4
+        with:
+          package: kubetail
+          package_root: .debpkg
+          maintainer: Andres Amorey
```
amorey: Andres Morey

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+          version: ${{ steps.tagName.outputs.tag }}
+          arch: ${{ matrix.pkgArch }}
+          desc: cli tool to get realtime logs from kubernetes
```
amorey: "Real-time logging dashboard for Kubernetes"

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.23.4"
+          cache: false
+      - name: Run build script
+        run: make build VERSION=${{ steps.tagName.outputs.tag }}
+
+      - name: create dir structure for .deb package
+        run: |
+          mkdir -p .debpkg/usr/bin
```
amorey, line 120: Just wondering - why did you use a hidden directory here? Also, is "/usr/bin" the intended installation location after the user installs the package?

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+      - name: Run build script
+        run: make build VERSION=${{ steps.tagName.outputs.tag }}
+
+      - name: install the rpm and rpm build
+        run: sudo apt install rpm build-essential
+
+      - name: copy spec file to SPEC dir
+        run: |
+          mkdir -p ~/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
+          cp .github/workflows/kubetail-rpm.spec ~/rpmbuild/SPECS/kubetail.spec
```
amorey: Let's put the .spec file in a new directory at `.github/ci-config`

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+          tar czf "kubetail-${CLI_VER}.tar.gz" "kubetail-${CLI_VER}/"
+          ls -la
+
+      - name: build RPM package
+        run: |
+          CLI_VER='${{ steps.tagName.outputs.tag }}'
+          ARCH='${{ matrix.pkgArch }}'
+
+          if [[ "${ARCH}" == "x86_64" ]]; then
+            rpmbuild -ba ~/rpmbuild/SPECS/kubetail.spec --define "version ${CLI_VER}" --define "arch ${ARCH}"
```
amorey: `--define` is used twice. Is that ok?

.github/workflows/release-cli.yml
```diff
@@ -86,9 +93,151 @@ jobs:
...
+          cd ~/rpmbuild/SOURCES
+          tar czf "kubetail-${CLI_VER}.tar.gz" "kubetail-${CLI_VER}/"
+          ls -la
+
+      - name: build RPM package
+        run: |
+          CLI_VER='${{ steps.tagName.outputs.tag }}'
+          ARCH='${{ matrix.pkgArch }}'
+
+          if [[ "${ARCH}" == "x86_64" ]]; then
```
amorey: Is it not possible to combine both cases into one command?
```console
rpmbuild -ba --target "${ARCH}" ~/rpmbuild/SPECS/kubetail.spec --define "version ${CLI_VER}"
```

amorey:
Instead of re-building the CLI tool from scratch I would recommend downloading the appropriate artifacts from the "build" step and using them to build the rpm and deb packages instead.

.github/workflows/kubetail-rpm.spec
```diff
@@ -0,0 +1,23 @@
+Name:           kubetail 
+Version:        %{version} 
+Release:        1%{?dist}
+Summary:        Go based cli to get logs from kubernetes
```
amorey: "Real-time logging dashboard for Kubernetes"

.github/workflows/kubetail-rpm.spec
```diff
@@ -0,0 +1,23 @@
...
+Summary:        Go based cli to get logs from kubernetes 
+
+License:        Apache License Version 2.0
+URL:            https://github.com/kubetail-org/kubetail
+Source0:        %{name}-%{version}.tar.gz
+
+BuildArch:      %{arch} 
+
+%description
+A longer description of your Go CLI tool and what it does.
```
amorey: "Kubetail is a general-purpose logging dashboard for Kubernetes, optimized for tailing logs across across multi-container workloads in real-time. With Kubetail, you can view logs from all the containers in a workload (e.g. Deployment or DaemonSet) merged into a single, chronological timeline, delivered to your browser or terminal."

.github/workflows/kubetail-rpm.spec
```diff
@@ -0,0 +1,23 @@
+Name:           kubetail 
+Version:        %{version} 
+Release:        1%{?dist}
+Summary:        Go based cli to get logs from kubernetes 
+
+License:        Apache License Version 2.0
```
amorey: Usually this is written as "Apache-2.0"

.github/workflows/kubetail-rpm.spec
```diff
@@ -0,0 +1,23 @@
+Name:           kubetail 
+Version:        %{version} 
+Release:        1%{?dist}
```
amorey, line 3: What's the difference between version and release?

.github/workflows/kubetail-rpm.spec
```diff
@@ -0,0 +1,23 @@
...
+Summary:        Go based cli to get logs from kubernetes 
+
+License:        Apache License Version 2.0
+URL:            https://github.com/kubetail-org/kubetail
+Source0:        %{name}-%{version}.tar.gz
+
+BuildArch:      %{arch}
```
amorey: Is `BuildArch` necessary? ChatGPT says if you exclude it, `rpmbuild` will default to the `--target` argument.

.github/workflows/release-cli.yml
victorchrollo14: the example contained it, so just used that. we could use any folder name we want. 
and yes "/usr/bin" is the location where the binary get's installed.

.github/workflows/release-cli.yml
victorchrollo14: alright. we could use this for manifests we'll add for winget release ci as well.

.github/workflows/release-cli.yml
victorchrollo14: yes. that's  part of the syntax.

.github/workflows/release-cli.yml
victorchrollo14: I think we can remove the BuildArch inside kubetail.spec part and use --target for both, that should do it.

.github/workflows/kubetail-rpm.spec
victorchrollo14: ok got it, I just copied this from the license file we have on kubetail repo, will change it up

victorchrollo14:
> Instead of re-building the CLI tool from scratch I would recommend downloading the appropriate artifacts from the "build" step and using them to build the rpm and deb packages instead.

sounds good. that prevents  us from repeating the steps for building cli in 3 different places.

.github/workflows/kubetail-rpm.spec
victorchrollo14: will remove this up and use --target for both arch.

.github/workflows/release-cli.yml
amorey: Sounds good

.github/workflows/release-cli.yml
```diff
@@ -86,9 +91,140 @@ jobs:
           name: artifacts-${{ matrix.GOOS }}-${{ matrix.GOARCH }}
           path: bin/*
 
+  package-deb:
+    name: Build deb package
+    needs: build
+    strategy:
+      matrix:
+        include:
+          - GOOS: linux
```
amorey: I think you don't need to keep `GOARCH` and `pkgArch` separate since they have the same value. Also, let's use `os` and `arch` instead of `GOOS` and `GOARCH` since we're not in a go context anymore.

.github/workflows/release-cli.yml
```diff
@@ -86,9 +91,140 @@ jobs:
...
+          path: bin
+          pattern: artifacts-${{ matrix.GOOS }}-${{ matrix.GOARCH }}
+          merge-multiple: true
+
+      - name: create dir structure for .deb package
+        run: |
+          mkdir -p .debpkg/usr/bin
+
+          # when we download from deb, the binary should be named kubetail
+          mv 'bin/kubetail-${{ matrix.GOOS }}-${{ matrix.GOARCH }}' bin/kubetail
```
amorey: You can just copy from `bin/` directly into `.depbkg/usr/bin` and change the name in one step.

.github/workflows/release-cli.yml
```diff
@@ -86,9 +91,140 @@ jobs:
...
+          arch: ${{ matrix.pkgArch }}
+          desc: "Real-time logging dashboard for Kubernetes"
+
+      - name: Calculate sha256 checksum
+        run: |
+          FILE='./kubetail_${{ steps.tagName.outputs.tag }}_${{ matrix.pkgArch }}.deb'
+          OUTPUT_FILE=${FILE}.sha256
+          sha256sum "$FILE" | cut -d " " -f 1 >"$OUTPUT_FILE"
+          mkdir -p deb-output
+          mv ./kubetail* ./deb-output/
```
amorey: This line gets executed in the root of the repository so if we add any files that begin with "kubetail" in the future they will get uploaded too.

.github/workflows/release-cli.yml
```diff
@@ -86,9 +91,140 @@ jobs:
...
+          path: bin
+          pattern: artifacts-${{ matrix.GOOS }}-${{ matrix.GOARCH }}
+          merge-multiple: true
+
+      - name: Rename the build binary to kubetail
+        run: |
+          mv 'bin/kubetail-${{ matrix.GOOS }}-${{ matrix.GOARCH }}' bin/kubetail
+
+      - name: install the rpm and rpm build
+        run: sudo apt install rpm build-essential
```
amorey: Some suggestions:
- You should run `update` before installing things from the respository
- I think you need to use `install -y` to automatically approve downloads
- Doesn't the ubuntu runner use "apt-get"?
- I'm used to seeing dependency management come in earlier steps. I would do this before downloading artifacts

```console
sudo apt-get update
sudo apt-get install -y rpm build-essential
```

.github/workflows/release-cli.yml
```diff
@@ -86,9 +91,140 @@ jobs:
...
+          cp .github/ci-config/kubetail-rpm.spec ~/rpmbuild/SPECS/kubetail.spec
+
+      - name: create tarball for the binary
+        run: |
+          CLI_VER='${{ steps.tagName.outputs.tag }}'
+          mkdir -p ~/rpmbuild/SOURCES/kubetail-${CLI_VER}
+          mv bin/kubetail ~/rpmbuild/SOURCES/kubetail-${CLI_VER}
+          cd ~/rpmbuild/SOURCES
+          tar czf "kubetail-${CLI_VER}.tar.gz" "kubetail-${CLI_VER}/"
+          ls -la
```
amorey: Is this a debug statement?

victorchrollo14: sorry, forgot to remove that.

.github/workflows/release-cli.yml
victorchrollo14: will change this up. didn't know the diff between apt and apt-get, seems like apt-get is more stable and recommended to be used in scripts.

.github/workflows/release-cli.yml
victorchrollo14: that makes sense, and then we'll be wondering why random files are showing up on the release.

.github/workflows/release-cli.yml
amorey: As far as I know "apt-get" is for Ubuntu and "apt" is for Alpine but maybe the Ubuntu runners aliased "apt"

victorchrollo14:
latest tested action for this: https://github.com/victorchrollo14/kubetail/actions/runs/15208648477 and the release https://github.com/victorchrollo14/kubetail/releases/tag/untagged-826076264e18beea8e9b

PR #335, opened 2025-05-24
🐋  add a workflow to publish kubetail windows binary to winget repository

### LINKED ISSUE #286: Push new releases to winget
Recently we added the latest version of `kubetail` to the winget repository (cli/v0.5.1) but winget doesn't pick up on new versions automatically. We should add a GitHub workflow (or modify the existing CLI release workflow) to push releases to winget.

victorchrollo14:
Fixes #286 

## Summary

<-- Explain the goal of your PR  -->
Adds a new ci workflow to automatically publish the latest kubetail windows binary to winget repository, whenever we publish a release.

## Changes

<-- Explain the changes in your PR -->
add  a new ci file called publish-winget and added the manifest templates under manifests folder.
we use the manifest template and update them based on the release version and raise a PR on the winget-pkgs repository.

## Submitter checklist

- [x] Rebase branch to HEAD


.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          sparse-checkout: .github/manifests
+          sparse-checkout-cone-mode: false
+          fetch-depth: 1
+          path: winget-manifests
+
+      - name: Clone winget-pkgs forked repo
+        uses: actions/checkout@v4
+        with:
+          repository: victorchrollo14/winget-pkgs
```
rxinui: I would recommend to transfer the ownership to `kubetail-org` as it make more sense to not be collaborator dependent.

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+  winget-release:
+    name: Release on winget
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Download sha256 files
+        id: release-info
+        uses: robinraju/release-downloader@v1
+        with:
+          latest: true
+          fileName: "kubetail-windows*zip.sha256"
```
rxinui, line 18: What kind of meta information do you need?

Because of the event type `on.release.types: publish` you already have some information from `github` context such as `github.sha` and this https://docs.github.com/en/webhooks/webhook-events-and-payloads#release

Perhaps this can complement https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#get-a-release-asset

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+          TAG_NAME="${{ steps.release-info.outputs.tag_name }}"
+
+          echo "tag=${TAG_NAME#cli/v}" >> $GITHUB_OUTPUT
+          echo "amd64_sha=$AMD64_SHA" >> $GITHUB_OUTPUT
+          echo "arm64_sha=$ARM64_SHA" >> $GITHUB_OUTPUT
+
+      - name: Download yq package
+        run: |
+          sudo wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/local/bin/yq 
+          sudo chmod +x /usr/local/bin/yq
```
rxinui: I would be better to stay at a github actions level, hence making use of

https://github.com/marketplace/actions/yq-portable-yaml-processor#github-action

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Update manifests for latest release
+        run: |
+          TAG_NAME='${{ steps.meta.outputs.tag }}'
+          BASE_PATH=https://github.com/kubetail-org/kubetail/releases/download
+          AMD64_SHA=${{ steps.meta.outputs.amd64_sha }}
+          ARM64_SHA=${{ steps.meta.outputs.arm64_sha }}
+
+          cd ./winget-manifests/.github/manifests
```
rxinui: Use this https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/setting-a-default-shell-and-working-directory#example-set-the-default-shell-and-working-directory

instead of `cd`

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+          echo "tag=${TAG_NAME#cli/v}" >> $GITHUB_OUTPUT
+          echo "amd64_sha=$AMD64_SHA" >> $GITHUB_OUTPUT
+          echo "arm64_sha=$ARM64_SHA" >> $GITHUB_OUTPUT
+
+      - name: Download yq package
+        run: |
+          sudo wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/local/bin/yq 
+          sudo chmod +x /usr/local/bin/yq
+
+      - name: Update manifests for latest release
```
rxinui: I would generalise by the mean of `strategy.matrix`. It is a good use case for multi-arch.

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+          repository: victorchrollo14/winget-pkgs
+          token: ${{ secrets.GH_TOKEN }}
+          path: winget-pkgs
+
+      - name: Extract tags and sha
+        id: meta
+        run: |
+          AMD64_SHA=$( tr -d '\r' < './kubetail-windows-amd64.zip.sha256' )
+          ARM64_SHA=$( tr -d '\r' < './kubetail-windows-arm64.zip.sha256' )
+          TAG_NAME="${{ steps.release-info.outputs.tag_name }}"
```
rxinui, line 39: you checked out the release therefore you can use `git describe --tags --abbrev=0`

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.installer.yaml 
+          yq e ".Installers[0].InstallerUrl = \"${BASE_PATH}/cli%2Fv${TAG_NAME}/kubetail-windows-amd64.zip\"" -i ./Kubetail.Kubetail.installer.yaml
+          yq e ".Installers[1].InstallerUrl = \"${BASE_PATH}/cli%2Fv${TAG_NAME}/kubetail-windows-arm64.zip\"" -i ./Kubetail.Kubetail.installer.yaml
+
+          yq e ".Installers[0].InstallerSha256 = \"${AMD64_SHA}\"" -i ./Kubetail.Kubetail.installer.yaml
+          yq e ".Installers[1].InstallerSha256 = \"${ARM64_SHA}\"" -i ./Kubetail.Kubetail.installer.yaml
+
+          # update the locale and kubetail yaml 
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.locale.en-US.yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.yaml
```
rxinui: I would prefer to use https://mikefarah.gitbook.io/yq/usage/github-action

and make usage of the `strategy.matrix`. More generic and powerful when we will add a new arch.

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GH_TOKEN }}
+          path: ./winget-pkgs
+          commit-message: "chore: Add manifest for Kubetail ${{ steps.meta.outputs.tag }}"
+          title: "New version: Kubetail.Kubetail version ${{ steps.meta.outputs.tag }}"
+          body: "This PR adds the manifest for Kubetail.Kubetail version ${{ steps.meta.outputs.tag }}."
+          branch: winget-release-${{ steps.meta.outputs.tag }}
+          base: master
+          push-to-fork: victorchrollo14/winget-pkgs
```
rxinui: Change the ownership i guesss

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          NEW_FOLDER='./winget-pkgs/manifests/k/Kubetail/Kubetail/${{ steps.meta.outputs.tag }}'
+          mkdir -p "$NEW_FOLDER" 
+
+          cp ./winget-manifests/.github/manifests/* "$NEW_FOLDER"/
+
+      - name: Commit and raise PR
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GH_TOKEN }}
```
rxinui: `secrets.GITHUB_TOKEN` should suffice?

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,89 @@
...
+          # update the locale and kubetail yaml 
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.locale.en-US.yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.yaml
+
+      - name: Copy manifests to winget-pkgs
+        run: |
+          NEW_FOLDER='./winget-pkgs/manifests/k/Kubetail/Kubetail/${{ steps.meta.outputs.tag }}'
+          mkdir -p "$NEW_FOLDER" 
+
+          cp ./winget-manifests/.github/manifests/* "$NEW_FOLDER"/
```
rxinui: what is that about?

rxinui (changes requested):
What is the need to have `.github/manifests/` folder?

.github/workflows/publish-winget.yml
amorey: We can use this: https://github.com/kubetail-org/winget-pkgs

.github/workflows/publish-winget.yml
victorchrollo14: we need the sha of the windows arm64 and amd64 zip files, which is uploaded to the assets during release.
I did think of making api request to do so, but this action seemed to keep it clean, so used this one.

along with that we just need the release tag and extract the version number from it.

.github/workflows/publish-winget.yml
victorchrollo14: sure. but I need to see if can run multiple yq commands using this action. will test it and see.

.github/workflows/publish-winget.yml
victorchrollo14: thanks. didn't know this existed.

.github/workflows/publish-winget.yml
victorchrollo14: I don't think matrix strategy works here, since we need to update a single yaml file with both arch at once and raise a PR. ( Assumption: matrix stratergy would run 2 jobs here for amd64 and arm64 )

.github/workflows/publish-winget.yml
victorchrollo14: github didn't allow me to create a secret which starts with "GITHUB"

.github/workflows/publish-winget.yml
victorchrollo14: we updated the original manifests template from .github/manifests in the above step,  then we copy those to a folder under our kubetail folder in winget-pkgs and then we raise a pr to winget-pkgs to make the release.

victorchrollo14:
> Re: .github/manifests
> 
> We added a `.github/ci-config` folder in a previous PR for the rpm spec. ChatGPT suggested this as the preferred way to share config-type files with actions as opposed to hard-coding docs inside the actions themselves and it sounded reasonable to me. We could continue use that folder and add a winget sub-directory or rename it.

sure, can we do something like .github/ci-config/manifests so we keep the winget manifests under a seperate folder.

.github/workflows/publish-winget.yml
victorchrollo14: example pr: https://github.com/victorchrollo14/winget-pkgs/pull/1

amorey:
> sure, can we do something like .github/ci-config/manifests so we keep the winget manifests under a seperate folder.

How about `.github/ci-config/winget` so it's clear that all the files are for winget?

amorey:
The test failure is from a race condition that should be fixed in `main`. Try rebasing and push again.

.github/workflows/publish-winget.yml
victorchrollo14: can't run multiple yq commands at once with this action, so sticking with current approach only.

.github/workflows/publish-winget.yml
amorey: `secrets.GITHUB_TOKEN` is included in the environment automatically

.github/ci-config/winget/Kubetail.Kubetail.installer.yaml
```diff
@@ -0,0 +1,19 @@
+# Created with YamlCreate.ps1 Dumplings Mod
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.10.0.schema.json
+
+PackageIdentifier: Kubetail.Kubetail
+PackageVersion: 0.5.1
```
amorey: Since these are templates let's replace version specific info with `<placeholder>` or something else

rxinui: Agree or empty string

.github/ci-config/winget/Kubetail.Kubetail.installer.yaml
```diff
@@ -0,0 +1,19 @@
...
+    PortableCommandAlias: kubetail
+Installers:
+  - Architecture: x64
+    InstallerUrl: https://github.com/kubetail-org/kubetail/releases/download/cli%2Fv0.5.1/kubetail-windows-amd64.zip
+    InstallerSha256: 2AF8F5107FBB9DAD521E1A7CB883560700D6B9353F2E4E125B1E803BF04AA332
+  - Architecture: arm64
+    InstallerUrl: https://github.com/kubetail-org/kubetail/releases/download/cli%2Fv0.5.1/kubetail-windows-arm64.zip
+    InstallerSha256: FB45771E1E91C9ADABBD763F0D3A23DE456569DE76E45F2BA54A0CF7B89B490E
+ManifestType: installer
+ManifestVersion: 1.10.0
```
rxinui: @amorey i would see these files as templates better.

We could use gotemplate or jinja template, to render it on CI as anything within `.github/` is CI related.

It makes the intent clearer

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,97 @@
...
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.installer.yaml
+          yq e ".Installers[0].InstallerUrl = \"${BASE_PATH}/cli%2Fv${TAG_NAME}/kubetail-windows-amd64.zip\"" -i ./Kubetail.Kubetail.installer.yaml
+          yq e ".Installers[1].InstallerUrl = \"${BASE_PATH}/cli%2Fv${TAG_NAME}/kubetail-windows-arm64.zip\"" -i ./Kubetail.Kubetail.installer.yaml
+
+          yq e ".Installers[0].InstallerSha256 = \"${AMD64_SHA}\"" -i ./Kubetail.Kubetail.installer.yaml
+          yq e ".Installers[1].InstallerSha256 = \"${ARM64_SHA}\"" -i ./Kubetail.Kubetail.installer.yaml
+
+          # update the locale and kubetail yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.locale.en-US.yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.yaml
```
rxinui, line 75: Although it is correct, I think it is a bit verbose.

We could have a follow up PR to improve the workflow and leverage a template renderer instead of manually update fields.

NOT A BLOCKER.

.github/workflows/publish-winget.yml
rxinui: By default, each run, GITHUB_TOKEN is generated.

You might need to explicitly define `permissions:` in your workflow to grand and restrict some access

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,97 @@
...
+          path: winget-pkgs
+
+      - name: Sync fork with upstream
+        working-directory: winget-pkgs
+        run: |
+          git remote add upstream https://github.com/microsoft/winget-pkgs.git
+          git fetch upstream master
+          git checkout master
+          git rebase upstream/master
+          git push origin master --force
```
rxinui: This is not robust. A `git rebase upstream/master` can encounter conflict which will result the workflow as error.

I don't see the point anyway to make use of a fork. We should remove all fork related work. Everything should be under `kubetail-org`

CC: @amorey

rxinui (changes requested):
Few details to take care of

.github/workflows/publish-winget.yml
amorey: Separate PR to turn these into templates sounds good

.github/workflows/publish-winget.yml
amorey: `git rebase upstream/master` will definitely encounter conflicts because the upstream `master` will get a new SHA when our PR is merged.

How can we issue a PR without maintaining our own fork?

If a fork is needed you can get around the conflict by creating a feature branch based on upstream/master:
```
git remote add upstream https://github.com/microsoft/winget-pkgs.git
git switch -c <branch-name> upstream/master
```

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,97 @@
...
+          # update the locale and kubetail yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.locale.en-US.yaml
+          yq e ".PackageVersion = \"$TAG_NAME\"" -i ./Kubetail.Kubetail.yaml
+
+      - name: Commit and raise PR
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GH_TOKEN }}
+          path: ./winget-pkgs
+          commit-message: "chore: Add manifest for Kubetail ${{ steps.meta.outputs.tag }}"
```
amorey: Looking at the PR's most of the commit messages are of the form "New version: Kubetail.Kubetail version <version>"

.github/workflows/publish-winget.yml
rxinui: I don't thin we need a fork. The reason why @victorchrollo14 has a fork is to test it out since he can't test on the upstream

victorchrollo14: we need a fork, we can't raise a pr to upstream directly since we don't have permissions to that repo, right now we have the write permissions to our fork at kubetail-org/winget-pkgs, hence we are able to push code and also raise PR.

victorchrollo14: can we try this.
1. clone the upstream directly.
2. checkout branch and make changes.
3. push to our forked repo and raise PR.   ( Will this work )

amorey: > we need a fork, we can't raise a pr to upstream directly since we don't have permissions to that repo, right now we have the write permissions to our fork at kubetail-org/winget-pkgs, hence we are able to push code and also raise PR.

This was my understanding too but I don't have any experience with this so maybe there's another way to create a PR? @rxinui

.github/workflows/publish-winget.yml
victorchrollo14: I did think of the template approach, but since we were updating only 5 values. I thought few commands would be fine. but let's change it up in an upcoming PR. for now I'll change to empty strings.

.github/workflows/publish-winget.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.PUBLISH_WINGET_TOKEN }}
+          path: ./winget-pkgs
+          commit-message: "New version: Kubetail.Kubetail ${{ steps.meta.outputs.tag }}"
+          title: "New version: Kubetail.Kubetail version ${{ steps.meta.outputs.tag }}"
+          body: "This PR adds the manifest for Kubetail.Kubetail version ${{ steps.meta.outputs.tag }}."
+          branch: winget-release-${{ steps.meta.outputs.tag }}
+          base: master
+          push-to-fork: victorchrollo14/winget-pkgs
```
rxinui: im still confuse about it.

Why push to fork? @amorey

rxinui (changes requested):

.github/workflows/publish-winget.yml
amorey: @victorchrollo14 Did you forget to change this to `push-to-fork: kubetail-org/winget-pkgs`?

The goal is to create a PR in the `microsoft/winget-pkgs` repo and, as external contributors, our PR needs to originate from a fork which in our case is `kubetail-org/winget-pkgs` (correct me if I'm wrong about the external contributors requirement).

The attributes in this action are confusingly named but @victorchrollo14 already tested it with his own fork and it worked. After @victorchrollo14 fixes the repo name the final result will be a PR requesting to merge `kubetail-org/winget-pkgs:winget-release-${{ steps.meta.outputs.tag }}` into `microsoft/winget-pkgs:master`.

victorchrollo14: > im still confuse about it.
> 
> Why push to fork? @amorey

it kind of works like this, we clone the upstream of microsoft/winget-pkgs and the make the pr through our fork branch. we need to change the target repo's before merging. if we don't have any more changes to be made

victorchrollo14: @amorey and @rxinui . if there are no further changes then we can change the clone repo to "microsoft/winget-pkgs" and  PR repo to "kubetail-org/winget-pkgs" and merge the PR.

amorey (dismissed):

victorchrollo14:
@amorey. squashed the commit's into one. changed the target repo's. we can merge this.

amorey:
👍 Let's wait for @rxinui then I'll merge

rxinui:
@victorchrollo14 you did a great job! This reviews are there not to annoy anyone but ensuring our flow is frictionless less, hope it's understood 😅🔥

victorchrollo14:
> @victorchrollo14 you did a great job! This reviews are there not to annoy anyone but ensuring our flow is frictionless less, hope it's understood 😅🔥

yeah. thanks for the reviews. I made a silly assumption which got weeded out during the review.

Issue #334, opened 2025-05-24
Upgrade pnpm to latest version (10.11.0)


amorey:
At some point in the pnpm release cycle, the runners on GitHub stopped being loaded with that version so we stopped keeping up-to-date. I believe this problem has been fixed so we should be able to upgrade now.

PR #340, opened 2025-05-25
fix(ci): Update pnpm/action-setup in ci.yml 

krisn2:
Updated the pnpm/action-setup version in ci.yml as suggested in issue #334.

### DIFF
--- a/.github/ci-config/kubetail-rpm.spec
+++ b/.github/ci-config/kubetail-rpm.spec
@@ -1,0 +1,27 @@
+Name:           kubetail 
+Version:        %{version} 
+Release:        1%{?dist}
+Summary:        Real-time logging dashboard for Kubernetes 
+
+License:        Apache-2.0
+URL:            https://github.com/kubetail-org/kubetail
+Source0:        %{name}-%{version}.tar.gz
+
+
+%description
+Kubetail is a general-purpose logging dashboard for Kubernetes,
+optimized for tailing logs across across multi-container workloads
+in real-time. With Kubetail, you can view logs from all the containers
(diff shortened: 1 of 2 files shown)


amorey:
It looks like the tests are failing because the docker image we're using can't install the latest version of pnpm.

Change the image tag to "22.16.0-alpine3.21" here:
https://github.com/kubetail-org/kubetail/blob/main/build/package/Dockerfile.dashboard#L17

Also, let's add the shasum here (sha512.6540583f41cc5f628eb3d9773ecee802f4f9ef9923cc45b69890fb47991d4b092964694ec3a4f738a420c918a333062c8b925d312f42e4f0c263eb603551f977): https://github.com/kubetail-org/kubetail/blob/main/dashboard-ui/package.json#L82

.github/workflows/ci.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 22
-      - uses: pnpm/action-setup@v3
+      - uses: pnpm/action-setup@v4
         with:
           version: 9
```
amorey: One last small change, let's set pnpm to `version: 10` everywhere

.github/workflows/release-cli.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 22
-      - uses: pnpm/action-setup@v3
+      - uses: pnpm/action-setup@v4
         with:
           version: 9
```
amorey: `version: 10` here too

amorey (approved):
/lgtm

PR #348, opened 2025-05-28
🎣 Fix uninitialized config bug and refactor kubeconfig handling code

### LINKED ISSUE #225: Add support for KUBECONFIG env var

amorey:
Fixes #225

## Summary

This PR reverts to @bavarianbidi original strategy to add support for KUBECONFIG env var by utilitzing the Kubernetes clientcmd library at config generation time, and in the process solves an uninitialized kube config bug in the previous commit. As where in the previous commit, we were doing `--kubeconfig` flag and `KUBECONFIG` env handling in the command loop, this PR offloads the loading logic to the Kubernetes clientcmd library at the time the kube config object is needed.

## Changes

- Removes use of default value for `--kubeconfig` value (now default is "")
- Switches to use of `kubeconfigPath` and `WithKubeconfigPath()` where appropriate to avoid confusion with kube config object instances
- Command loops pass through value of `--kubeconfig` to lower level code that initializes kube config objects when needed
- Modifies CLI `NewTunnel()` method (modules/cli/internal/tunnel/tunnel.go) to use Kubernetes clientcmd library to initialize config so `KUBECONFIG` env var is handled automatically
- Modifies `KubeConfigWatcher` to use Kubernetes clientcmd library to obtain list of config files from `--kubeconfig` value and `KUBECONFIG` env var but currently only utilizes the first file rather than all the files
- Temporarily disables flaky `KubeConfigWatcher` tests
- Whitelists "@swc/core", "esbuild" in pnpm build instructions to avoid unrelated dashboard interactive build issue

Notes:
- We need to add support for multiple files to `KubeConfigWatcher` in a followup PR
- We need to add back `KubeConfigWatcher` async tests and fix flakiness issue

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/config/config_test.go
+++ b/modules/shared/config/config_test.go
@@ -3,11 +3,14 @@ package config
 import (
 	"net/http"
 	"os"
+	"path/filepath"
+	"reflect"
 	"testing"
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 var cfg1 = `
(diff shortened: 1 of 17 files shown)


bavarianbidi (approved):
lgtm

I hope this implementation feels better to you and your initial concerns are gone :crossed_fingers:

modules/shared/config/config_test.go
```diff
@@ -32,3 +33,28 @@ func TestConfig(t *testing.T) {
 	assert.Equal(t, EnvironmentCluster, cfg.Dashboard.Environment)
 	assert.Equal(t, http.SameSiteStrictMode, cfg.Dashboard.Session.Cookie.SameSite)
 }
+
+func TestNewConfig(t *testing.T) {
```
rxinui, line 37: That is great!

modules/shared/helm/helm.go
```diff
@@ -389,10 +389,10 @@ func NewClient(options ...ClientOption) *Client {
 	return c
 }
 
-// Option Kubeconfig
-func WithKubeconfig(kubeconfig string) ClientOption {
+// Option KubeconfigPath
+func WithKubeconfigPath(kubeconfigPath string) ClientOption {
 	return func(c *Client) {
-		c.KubeConfig = kubeconfig
+		c.KubeConfig = kubeconfigPath
```
rxinui: Isn't it a bit confusing to have two different namings?

not a blocker.

modules/shared/k8shelpers/kube-config-watcher.go
```diff
@@ -69,17 +60,17 @@ func NewKubeConfigWatcher(kubeconfigPath string) (*KubeConfigWatcher, error) {
...
 		return nil, err
 	}
 
 	// Initialize
 	w := &KubeConfigWatcher{
-		kubeconfigPath: kubeconfigPath,
-		kubeConfig:     kubeConfig,
-		watcher:        watcher,
-		eventbus:       evbus.New(),
+		kubeConfig: kubeConfig,
```
rxinui, line 71: That goes back to my command. I would favour `kubeConfig` as a instance of config and `KubeconfigPath` as a filepath.

modules/shared/k8shelpers/kube-config-watcher_test.go
```diff
@@ -140,122 +139,122 @@ func TestKubeConfigWatcherSubscribeAdded(t *testing.T) {
...
-
-	// Delete file
-	err = os.Remove(kubeconfigPath)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	wg.Wait()
-}
+// func TestKubeConfigWatcherSubscribeModified(t *testing.T) {
```
rxinui: Delete it if not use.

rxinui (changes requested):
Minor adjustments

modules/shared/config/config_test.go
amorey: Credit goes to @bavarianbidi

modules/shared/helm/helm.go
amorey: Agreed, but `KubeConfig` attribute name comes from helm `*cli.EnvSettings` which is embedded in the `Client`. To make things more transparent we could revert to attaching `*cli.EnvSettings` as an attribute or set the attribute through `EnvSettings` explicitly:

```go
c.EnvSettings.KubeConfig = kubeconfigPath
```

What do you think?

modules/shared/k8shelpers/kube-config-watcher_test.go
amorey: Ok, removed. Will add it back in a follow-up pr.

modules/shared/helm/helm.go
rxinui: `c.EnvSettings.KubeConfig = kubeconfigPath` seems explicit and Go aims to be exactly that.

I would favour this, however if this implies a lot of code change, i would rather not go down that road...

amorey: Done

PR #364, opened 2025-05-31
🎣 Add worker to cache only after successful start in DesktopHealthMonitor

misshanya:
## Summary

Previously, we added the worker to the cache before starting it. If start failed, then non-working worker would still remain in cache, which could lead to returning a broken instance later. I don't think it is right behavior.

Now, we only store the worker in the cache after it has successfully started

## Changes

- Moved caching to occur only after successful start

## Submitter checklist

- [x] Rebase branch to HEAD


amorey (approved):

amorey:
Nice catch

PR #368, opened 2025-06-01
🐋 Adding OpenTelemetry package and instrumentation to StreamForward in cluster-agent

Salazar-99:
⚠️ Work In Progress

## Summary
To begin the process of instrumenting the codebase with OpenTelemetry, this PR adds some initial infrastructure and instruments a single handler in `cluster-api`.

## Changes
- Added the `otel` package in the `shared` module. 
- Added `OTel` to the `config` package in the `shared` module.
- Added instrumentation code in `cluster-agent` to wrap the gRPC server and added a span to the `StreamForward` method. 

## Demo
[Short Demo Video](https://drive.google.com/file/d/1SogHO-0SZlXR-WE1-_eRshq2Fzappmfs/view?usp=sharing)
This video shows me triggering the `StreamForward` method from the KubeTail UI and subsequently seeing the corresponding span in the HyperDX (ClickStack) UI.

## Submitter checklist

- [ ] Add summary and explain changes in the PR description
- [ ] Rebase branch to HEAD


hack/tilt/kubetail.yaml
```diff
@@ -430,6 +430,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: OTEL_EXPORTER_OTLP_HEADERS
+          value: "Authorization=<get-api-key-from-hyperdx-ui>"
```
amorey: Does HyperDX require you to obtain a new API key every time the app is installed? It would be convenient if we could hard code an API key here or disable authentication altogether.

hack/tilt/kubetail.yaml
```diff
@@ -500,6 +502,13 @@ spec:
         matchLabels:
           app.kubernetes.io/name: kubetail
           app.kubernetes.io/component: cluster-api
+  - from:
```
amorey: Does clickstack require two-way communication?

modules/shared/config/config.go
```diff
@@ -56,6 +56,15 @@ type Config struct {
 	AllowedNamespaces []string `mapstructure:"allowed-namespaces"`
 	KubeconfigPath    string   `mapstructure:"kubeconfig"`
 
+	// OTel options
+	OTel struct {
```
amorey: I think each component (Dashboard, Cluster API, Cluster Agent) should have its own instance of the OTel options so we can keep them independent. Soon we will probably split up the shared config into separate configs so having separate otel configs will help that.

modules/shared/otel/otel.go
```diff
@@ -0,0 +1,87 @@
...
+		// creating spans and no telemetry will be collected or sent
+		otel.SetTracerProvider(noop.NewTracerProvider())
+		return nil
+	}
+
+	// Define common attributes for all spans
+	resources, err := resource.New(
+		context.Background(),
+		resource.WithAttributes(
+			attribute.String("service.name", cfg.ServiceName),
```
amorey, line 42: Is "service.name" standard in otel? "service" is a bit confusing because the name overlaps with Kubernetes `Service` resources and also gRPC "services". How about using Kubernetes label-like names (app.name, app.component, etc.)?

modules/shared/config/config.go
```diff
@@ -236,6 +245,12 @@ func (cfg *Config) validate() error {
 func DefaultConfig() *Config {
 	cfg := &Config{}
 
+	// TODO: Bubble this config up somewhere closer to main
+	cfg.OTel.Enabled = true
```
amorey: OTel should be disabled by default. Otherwise users will have to run an otel receiver.

modules/shared/config/config.go
```diff
@@ -236,6 +245,12 @@ func (cfg *Config) validate() error {
...
+	cfg.OTel.Debug = true
```
amorey: I think Debug should be `false` by default. That way if/when users enable it in production they only have to set `Enabled` to `true`.

modules/cluster-agent/internal/services/logrecords/logrecords.go
```diff
@@ -57,28 +66,47 @@ func (s *LogRecordsService) Shutdown() {
 
 // Implementation of StreamForward() in LogRecordsService
 func (s *LogRecordsService) StreamForward(req *clusteragentpb.LogRecordsStreamRequest, stream clusteragentpb.LogRecordsService_StreamForwardServer) error {
+	ctx, span := tracer.Start(stream.Context(), "logrecords.StreamForward",
+		trace.WithSpanKind(trace.SpanKindServer),
+		trace.WithAttributes(
```
amorey, line 71: Is there a way to send the entire `clusteragentpb.LogRecordsStreamRequest` instance as json or something else?

modules/shared/otel/otel.go
```diff
@@ -0,0 +1,87 @@
...
+				trace.WithBatchTimeout(time.Second)),
+			trace.WithResource(resources))
+		otel.SetTracerProvider(traceProvider)
+		return nil
+	}
+
+	// Development mode use gRPC trace exporter to an OTel collector without TLS
+	// since this is a local development setup only for now.
+	var traceExporter trace.SpanExporter
+	if cfg.Endpoint != "" {
```
amorey: Can we simplify this block by using `WithEndpoint()` for all values of `Endpoint`?
```go
traceExporter, err = otlptracegrpc.New(
  context.Background(),
  otlptracegrpc.WithInsecure(),
  otlptracegrpc.WithEndpoint(cfg.Endpoint),
)
```

PR #406, opened 2025-06-06
🎣 feat: Add Sec-Fetch-Site validation for WebSocket connections

### LINKED ISSUE #404: Add `sec-fetch-site` header based CSRF protection to Dashboard server
Until now, we've been using cookie-based CSRF protections in the Dashboard and Cluster API to protect against cross-site WebSocket requests. This solution isn't ideal because we have to send the CSRF token in a separate WebSocket authorization payload. Also, it makes using third-party tools to query our GraphQL endpoints more difficult because we have to add custom code to perform CSRF auth.

As a first step in replacing our cookie-based CSRF protections with something simpler we would like to deploy `sec-fetch-site` header-based protections. Doing this will let us test our assumptions about our users' browsers (i.e. that they're "modern") in a way that's easy to roll back if we run into problems.

ramsyana:
Fixes #404 

## Summary

This PR adds an additional layer of security to our WebSocket connections by validating the Sec-Fetch-Site header, as we discussed. This change hardens our defense against Cross-Site WebSocket Hijacking (CSWSH) while maintaining compatibility for clients that don't send the header.

## Changes

- The CheckOrigin function in both the Dashboard and Cluster API servers has been updated.
- It now checks if the Sec-Fetch-Site header is same-origin or same-site.
- If the header is not present, it gracefully falls back to the existing cookie-based CSRF validation in the InitFunc.
- This change is purely additive and does not remove any of the existing CSRF protections, making it a safe first step.

## Submitter checklist

- ✅ Add the correct emoji to the PR title
- ✅  Link the issue number, if any, to *Fixes #*
- ✅  Add summary and explain changes in the PR description
- ✅  Rebase branch to HEAD
- ✅  Squash changes into one signed, single commit

Thanks!

### DIFF
--- a/modules/cluster-api/graph/server.go
+++ b/modules/cluster-api/graph/server.go
@@ -48,7 +48,7 @@ type Server struct {
 
 // allowedSecFetchSite defines the secure values for the Sec-Fetch-Site header.
 // It's defined at the package level to avoid re-allocation on every WebSocket upgrade request.
-var allowedSecFetchSite = []string{"same-origin", "same-site"}
+var allowedSecFetchSite = []string{"same-origin"}
 
 // Create new Server instance
 func NewServer(cm k8shelpers.ConnectionManager, grpcDispatcher *grpcdispatcher.Dispatcher, allowedNamespaces []string, csrfProtectMiddleware func(http.Handler) http.Handler) *Server {
--- a/modules/cluster-api/graph/server.go
+++ b/modules/cluster-api/graph/server.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"time"
+	"slices"
 
 	"github.com/99designs/gqlgen/graphql/handler"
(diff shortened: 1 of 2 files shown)


modules/cluster-api/graph/server.go
```diff
@@ -45,6 +46,10 @@ type Server struct {
 	shutdownCh chan struct{}
 }
 
+// allowedSecFetchSite defines the secure values for the Sec-Fetch-Site header.
+// It's defined at the package level to avoid re-allocation on every WebSocket upgrade request.
+var allowedSecFetchSite = []string{"same-origin", "same-site"}
```
amorey: Should we only allow "same-origin"? Otherwise it will accept requests from other localhost ports.

ramsyana: Yeah. Will restrict to `"same-origin"` only to prevent localhost port hopping. Will update it shortly.

amorey (approved):

PR #410, opened 2025-06-07
🐋Add unit tests to cmd package in cli module

### LINKED ISSUE #409: Add unit tests to cmd in cli module
github.com/kubetail-org/kubetail/tree/main/modules/cli/cmd has no unit tests.

Unit tests should be added to ensure the server is initialized correctly and that each command is working as expected.

mochivi:
Fixes #409 

## Summary

This PR introduces unit tests to the [cmd](https://github.com/kubetail-org/kubetail/tree/main/modules/cli/cmd) package in the cli module. To make unit tests feasible, small changes to each file were made.

## Changes

Besides only adding unit tests, the strategy is that a --test flag was added to each tested command. This flag allows for verifying if the internal config is being created as expected.
For the cmd/serve.go file, the code that actually runs the server was extracted to a serve() function. This allowed the function to be tested in cmd/serve_test.go.

## Submitter checklist

- [x] Rebase branch to HEAD


modules/cli/cmd/serve.go
```diff
@@ -84,19 +84,16 @@ var serveCmd = &cobra.Command{
 			return
 		}
 
-		// listen for termination signals
```
amorey: We should listen for termination signals as early as possible. Is there a benefit to moving this code block lower?

amorey:
Curious to hear @rxinui thoughts on the testing pattern.

modules/cli/cmd/serve.go
mochivi: The quit channel that is intercepting the SIGTERM and SIGINT calls isn't used anywhere before this, therefore, I just moved the channel initialization closer to where it is actually used, the reason why it is even our of serve() is to allow for manual control of the channel in future tests.
SIGTERM and SIGINT will cancel the execution anyway at any point, the quit channel and signal.Notify() combination only really works to block the execution while it is serving and to prepare for graceful shutdown.

mochivi:
Need some help to understand why the test is failing when running here. Running the tests locally works just fine.

modules/cli/cmd/serve.go
amorey: AFAIK `signal.Notify()` doesn't start listening until you call it and if the os receives a signal before that point, then you'll miss it. This is why we start the listener when `Run()` is called even though `quit` it isn't used until later. Come to think of it, we should probably call it even before `Run()` but we can leave that for another time.

mochivi: From the signal package documentation, it looks like it will just override the default SIGTERM and SIGINT behavior. As I understand the below, the program will actually run for longer if we set up the signal.Notify earlier, as it will override the default behavior of SIGTERM and SIGINT.

[Changing the behavior of signals in Go programs](https://pkg.go.dev/os/signal#hdr-Changing_the_behavior_of_signals_in_Go_programs)

The functions in this package allow a program to change the way Go programs handle signals.

Notify disables the default behavior for a given set of asynchronous signals and instead delivers them over one or more registered channels. Specifically, it applies to the signals SIGHUP, SIGINT, SIGQUIT, SIGABRT, and SIGTERM. It also applies to the job control signals SIGTSTP, SIGTTIN, and SIGTTOU, in which case the system default behavior does not occur. It also applies to some signals that otherwise cause no action: SIGUSR1, SIGUSR2, SIGPIPE, SIGALRM, SIGCHLD, SIGCONT, SIGURG, SIGXCPU, SIGXFSZ, SIGVTALRM, SIGWINCH, SIGIO, SIGPWR, SIGINFO, SIGTHR, SIGWAITING, SIGLWP, SIGFREEZE, SIGTHAW, SIGLOST, SIGXRES, SIGJVM1, SIGJVM2, and any real time signals used on the system. Note that not all of these signals are available on all systems.

If the program was started with SIGHUP or SIGINT ignored, and [Notify](https://pkg.go.dev/os/signal#Notify) is called for either signal, a signal handler will be installed for that signal and it will no longer be ignored. If, later, [Reset](https://pkg.go.dev/os/signal#Reset) or [Ignore](https://pkg.go.dev/os/signal#Ignore) is called for that signal, or [Stop](https://pkg.go.dev/os/signal#Stop) is called on all channels passed to Notify for that signal, the signal will once again be ignored. Reset will restore the system default behavior for the signal, while Ignore will cause the system to ignore the signal entirely.

If the program is started with a non-empty signal mask, some signals will be explicitly unblocked as described above. If Notify is called for a blocked signal, it will be unblocked. If, later, Reset is called for that signal, or Stop is called on all channels passed to Notify for that signal, the signal will once again be blocked.

rxinui:
> Need some help to understand why the test is failing when running here. Running the tests locally works just fine.

I reckon this is due to https://github.com/kubetail-org/kubetail/actions/runs/15502196178/workflow?pr=410#L80

You should try `go test -race ./...` to test your local modules rather than those published. @mochivi

modules/cli/cmd/serve.go
amorey: We are using `signal.Notify()` to override the default behavior so we can implement a graceful shutdown.

mochivi:
> > Need some help to understand why the test is failing when running here. Running the tests locally works just fine.
> 
> I reckon this is due to https://github.com/kubetail-org/kubetail/actions/runs/15502196178/workflow?pr=410#L80
> 
> You should try `go test -race ./...` to test your local modules rather than those published. @mochivi

So, I did run the exact same command and it is successful locally, I am not sure why it is failing here.

PR #411, opened 2025-06-07
🐋  MCP Server

Bl4ck-h00d:
# 🐋  MCP Server

## Summary
This PR introduces the initial implementation of the Kubetail MCP (Model Context Protocol) server, enabling AI tools and automation clients to query Kubernetes logs using natural language and advanced filters. The MCP server leverages Kubetail’s existing log fetchers and supports infrastructure-aware filtering, providing a robust, AI-friendly log search and streaming interface.

## Changes

* Added a new `mcp-server` subcommand to the Kubetail CLI for MCP protocol support.
* Integrated the mark3labs/mcp-go library to handle MCP (Model Context Protocol) requests over stdio/HTTP.
* Implemented a three-tier log fetcher strategy:
     - Direct gRPC agent access (AgentLogFetcher)
     - Cluster API proxy via GraphQL (ClusterAPIProxyLogFetcher)
     - Kubernetes API fallback (KubeLogFetcher)
     
* Exposed a `kubernetes_logs_search` MCP tool with support for advanced filtering (namespace, pod, container, region, zone, arch, OS, node, grep, time range).
* Added support for multiple output formats: structured JSON, summary, and raw log lines.
* Included documentation and usage examples, including integration steps for Claude Desktop and other AI tools.
* Added utility functions for log summarization, keyword extraction, and metadata enrichment.

## Submitter checklist

- [x] Rebase branch to HEAD


modules/cli/cmd/mcp_server.go
```diff
@@ -0,0 +1,1254 @@
...
+	Since string `json:"since,omitempty" jsonschema:"description=Show logs since this time (ISO timestamp or duration like 'PT30M')"`
+	Until string `json:"until,omitempty" jsonschema:"description=Show logs until this time (ISO timestamp or duration like 'PT10M')"`
+
+	// Output control
+	Tail   int64 `json:"tail,omitempty" jsonschema:"description=Show last N log entries (default: 100)"`
+	Follow bool  `json:"follow,omitempty" jsonschema:"description=Stream new log entries in real-time"`
+}
+
+// Response structures for AI tool consumption
+type LogEntryResponse struct {
```
amorey: Did you consider using the `LogRecord` spec from https://github.com/kubetail-org/kubetail/blob/main/modules/shared/logs/stream.go#L31? In `LogRecord` we split out the `LogSource` object from the main record attributes (`timestamp`, `message`) so we can re-use it in the the log query API and also in the log source watcher.

modules/cli/cmd/mcp_server.go
```diff
@@ -0,0 +1,1254 @@
...
+		log.Debug().Err(err).Msg("Failed to create AgentLogFetcher, using KubeLogFetcher")
+		return logs.NewKubeLogFetcher(clientset), "KubeLogFetcher", nil
+	}
+
+	log.Info().Msgf("Using %s for server-side log processing", fetcherType)
+	return fetcher, fetcherType, nil
+}
+
+// tryCreateAgentLogFetcher attempts to create an AgentLogFetcher, falling back to ClusterAPIProxyLogFetcher, then KubeLogFetcher if unavailable.
+func tryCreateAgentLogFetcher(cm k8shelpers.ConnectionManager, kubeContext string, clientset kubernetes.Interface) (logs.LogFetcher, string, error) {
```
amorey: Do direct gRPC connections actually work? The k8s installation uses a NetworkPolicy that only allows traffic to the agents from the API server.

amorey:
I'm impressed at how you implemented this without any prior knowledge of the code base. I think I can help you polish it but I need more time to sit down to think about how best to do it. Unfortunately, this weekend got busy with family stuff but this is top priority so I'll get to it asap.

modules/cli/cmd/mcp_server.go
Bl4ck-h00d: I will update the `LogEntryResponse` to use `LogSource`, but I don't think we can use `LogRecord` because of differences in timestamp formatting and the internal error field.

modules/cli/cmd/mcp_server.go
Bl4ck-h00d: Yes, you're right. I had a misconception that the kubetail CLI might be running inside the Kubernetes cluster, so I assumed a direct connection might work. But I now realize users can connect remotely using their kubeconfig file, so that changes things.

I'll update the function.

amorey: No worries, you actually had the right instinct to use the `kubetail-cli` service account (again, very impressed).

For some background, `kubetail-cli` is a hack to allow the CLI tool to get a bearer token to use inside the cluster but the hack has reached its limit and I think we need to get rid of it in order to hande users with namespace-scoped permissions (https://github.com/orgs/kubetail-org/discussions/358). Currently part of the implementation strategy is to turn the Cluster API into a Kubernetes API extension which means in the future the CLI tool will connect to the Kubetail resources inside the cluster through the proper proxy Kubernetes mechanism, not through the `kubetail-cli` serviceaccount.

Do you think that in the future we should allow the MCP server to connect to the agents directly or should we use the "Cluster API" as a go-between? My assumption was that we would want to use the Cluster API always (to handle authentication, merge log streams server-side, etc.) but maybe my assumption is incorrect.

Bl4ck-h00d: I think using Cluster API as the middle layer makes a lot of sense. It keeps auth centralized and provides a consistent interface as we migrate to a Kubernetes API extension.

PR #412, opened 2025-06-08
🐋 Color and Shape added to container/pod list in dashboard-ui

### LINKED ISSUE #181: Improve container color coding algorithm
Currently, Kubetail uses [distinct-colors](https://github.com/internalfx/distinct-colors) to generate visually distinct colors for the colored dots that are used to differentiate between source containers. Initially 20 colors are chosen and containers are binned into them based on a hash of the pod+container name:
https://github.com/kubetail-org/kubetail/blob/main/dashboard-ui/src/pages/console.tsx#L68-L94

Unfortunately, sometimes the current implementation picks colors that are similar to each other and hard to distinguish on the screen (e.g. similar shades of blue or green). It would be great to improve the algorithm so that it's always easy to distinguish between colors.

One idea is to hand pick the colors (maybe using ANSI regular colors as a starting point). I think 10-20 colors should be fine more most users and we could provide an option for more/custom colors in the future.

Let's keep this issue open to get feedback on how to improve container color coding.

Nanashi-lab:
Fixes #181 

## Summary

Improve container color coding, use unique color palette for each of the two themes. 

## Changes

- Removed Distinct Color, No longer part of external packages
- Added 20 Unique Color for each theme and added 4 shapes (Circle, Square, Diamond, Triangle) [dashboard-ui/src/lib/color-shape.tsx](https://github.com/kubetail-org/kubetail/pull/412/files#diff-0d504ec9a47e2af90a16069ceb89a3ad83d9639e52fbe39cf63d856a2a4a0290)
- Created a new local storage, first 20 are circle with different colors, rest being randomly picked from rest of the shapes
- The colors smoothly change without needing a refresh if themes are changed
 
I will squash and push in a single commit once, all the changes are finalized. Same with Rebase  

## Submitter checklist

- [ ] Rebase branch to HEAD

Screenshot From 2025-06-08 20-23-16
Screenshot From 2025-06-08 20-23-26

### DIFF
--- a/dashboard-ui/package.json
+++ b/dashboard-ui/package.json
@@ -28,7 +28,6 @@
     "clsx": "^2.1.1",
     "date-fns": "^4.1.0",
     "date-fns-tz": "^3.2.0",
-    "distinct-colors": "^3.0.0",
     "fancy-ansi": "^0.1.3",
     "fast-deep-equal": "^3.1.3",
     "graphql": "^16.12.0",
--- a/dashboard-ui/pnpm-lock.yaml
+++ b/dashboard-ui/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
     dependencies:
       '@apollo/client':
         specifier: ^4.0.10
-        version: 4.0.10(graphql-ws@6.0.6(graphql@16.12.0)(ws@8.18.3))(graphql@16.12.0)(react-dom@19.2.1(react@19.2.1))(react@19.2.1)(rxjs@7.8.2)
+        version: 4.0.11(graphql-ws@6.0.6(graphql@16.12.0)(ws@8.18.3))(graphql@16.12.0)(react-dom@19.2.1(react@19.2.1))(react@19.2.1)(rxjs@7.8.2)
       '@fontsource-variable/roboto-flex':
         specifier: ^5.2.8
         version: 5.2.8
@@ -25,7 +25,7 @@ importers:
         version: 5.2.2(react-hook-form@7.68.0(react@19.2.1))
       '@kubetail/ui':
         specifier: 2.0.4
-        version: 2.0.4(cvr3v4d7miwoguvmpzrkq3nkba)
+        version: 2.0.4(f49e59bc6a8424a76672c78b49a410ae)
       '@radix-ui/react-checkbox':
         specifier: ^1.3.3
         version: 1.3.3(@types/react-dom@19.2.3(@types/react@19.2.7))(@types/react@19.2.7)(react-dom@19.2.1(react@19.2.1))(react@19.2.1)
@@ -62,9 +62,6 @@ importers:
       date-fns-tz:
         specifier: ^3.2.0
         version: 3.2.0(date-fns@4.1.0)
-      distinct-colors:
-        specifier: ^3.0.0
-        version: 3.0.0
       fancy-ansi:
         specifier: ^0.1.3
         version: 0.1.3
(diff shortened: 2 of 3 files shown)


dashboard-ui/src/lib/color-shape.tsx
```diff
@@ -0,0 +1,224 @@
...
+const getContainerOrder = (): string[] => {
+  try {
+    const stored = localStorage.getItem(CONTAINER_ORDER_STORAGE_KEY);
+    return stored ? JSON.parse(stored) : [];
+  } catch {
+    return [];
+  }
+};
+
+// Add container to order list if not already present
```
Nanashi-lab: I am unsure of having to create a new local storage key, the reason why we do this is, color uses a hash-based system on the hash of the container key (namespace/podName/containerName), his doesn't guarantee that the "first 20" containers will get circles.

 There might be a better way to do this. Or add a way to clear the local storage. Right now if I open a different cluster, it will treat each pod as new and assign them new shapes.

dashboard-ui/src/lib/hooks.ts
```diff
@@ -546,42 +546,6 @@ export function useLogMetadata(options?: LogMetadataHookOptions) {
   return { loading, error, data };
 }
 
-/**
```
Nanashi-lab: This useColors wasnt used any where, removed this, so distinct-color is no longer part of kubetail

dashboard-ui/src/lib/color-shape.tsx
```diff
@@ -0,0 +1,224 @@
...
+          style={baseStyle}
+        />
+      );
+  }
+};
+
+export { LIGHT_THEME_COLORS, DARK_THEME_COLORS };
+
+// Export utility function to clear container order (for testing/debugging)
+export const clearContainerOrder = (): void => {
```
gikaragia: I don't think this is needed in the code, for debugging we can delete the storage using the browser tools

dashboard-ui/src/lib/color-shape.tsx
```diff
@@ -0,0 +1,224 @@
...
+    // Use hash for consistent color assignment
+    const streamUTF8 = new TextEncoder().encode(key);
+    const buffer = await crypto.subtle.digest('SHA-256', streamUTF8);
+    const view = new DataView(buffer);
+    const colorSeed = view.getUint32(0);
+
+    let shapeIndex: number;
+    let colorIndex: number;
+
+    if (orderIndex < 20) {
```
gikaragia: Regardless if we drop randomness or not, this could go in a helper function

dashboard-ui/src/lib/color-shape.tsx
```diff
@@ -0,0 +1,224 @@
...
+  DIAMOND = 'diamond',
+  TRIANGLE = 'triangle',
+}
+
+const AVAILABLE_SHAPES = [Shape.CIRCLE, Shape.SQUARE, Shape.DIAMOND, Shape.TRIANGLE];
+
+// Local storage key for container order tracking
+const CONTAINER_ORDER_STORAGE_KEY = 'kubetail-container-order';
+
+const LIGHT_THEME_COLORS = [
```
gikaragia: Did you followed any guide to define the color values?

dashboard-ui/src/lib/color-shape.tsx
gikaragia: Generally, I think it would be great if the same pod was always represented by the same color.

gikaragia:
Hey @amorey so this is what has been discussed in this PR:

1. Keep the randomness and shapes: Each container's color and shape will be picked consistently by a hash of `{namespace}/{pod-name}/{container-name}` and we will support a maximum of 80 different containers.
2. Keep the randomness and drop the shapes: This is going to be the same as before but we will just have a static list of colors and we will support 20 different containers
3. Drop the randomness and keep the shapes: The first time a container is displayed, we calculate the next available combination of color and shape and store it for the pod in local storage. This way we support 80 different combinations.

Which approach do you think we should follow?

dashboard-ui/src/lib/color.tsx
```diff
@@ -0,0 +1,103 @@
...
+];
+
+export type ContainerColorProps = {
+  containerKey: string;
+  size?: 'small' | 'medium';
+  className?: string;
+  theme?: 'light' | 'dark';
+};
+
+export const ContainerColor = ({ containerKey, size = 'medium', className = '', theme = 'light' }: ContainerColorProps) => {
```
amorey: You can use the `useTheme()` hook here instead of passing `theme` as an argument.

PR #448, opened 2025-06-10
🎣  fix(api): prevent writing to hijacked WebSocket connections

### LINKED ISSUE #415: "Write on hijacked connection" error in dashboard server
A user reported seeing the following error message in the dashboard server logs:
```
2025/06/09 08:33:39 http: response.Write on hijacked connection from [github.com/gin-gonic/gin.(*responseWriter).Write](http://github.com/gin-gonic/gin.(*responseWriter).Write) (response_writer.go:83)
```

Here is Claude Code's initial diagnosis which sounds reasonable to me:
```
* This Gin error occurs when code tries to write to an HTTP response after the connection has been "hijacked" (typically
  for WebSocket upgrades).

  In kubetail's case, the issue is in your GraphQL subscription resolvers for log streaming:

  Problem locations:
  - modules/dashboard/graph/schema.resolvers.go:1041
  - modules/cluster-api/graph/schema.resolvers.go:309

  Root cause: The error handling code calls transport.AddSubscriptionError(ctx, gqlerrors.ErrInternalServerError) from
  goroutines that may run after the HTTP connection has been upgraded to WebSocket. Gin's response writer can't write to
  hijacked connections.

  Fix: Add connection state checks before error writes, or use the GraphQL subscription's proper error channels instead of
   trying to write HTTP responses.
```

ramsyana:
Fixes #415 

### Summary

This PR cleans up the `http: response.Write on hijacked connection` error that would occasionally appear in the server logs. This was happening when a GraphQL log subscription terminated, and our error handling tried to write to the connection after it was already upgraded to a WebSocket.

### Changes

-   In the `logRecordsFollow` subscription resolvers for both the `dashboard` and `cluster-api` modules, the call to `transport.AddSubscriptionError` has been replaced.
-   We now log the stream termination error directly on the server side. This correctly handles the error without causing a panic, and the client's subscription simply closes as expected.
-   Added the `zerolog` import to the dashboard's resolver to support this change.

### Submitter checklist

-   ✅ Add the correct emoji to the PR title
-   ✅ Link the issue number, if any, to *Fixes #*
-   ✅ Add summary and explain changes in the PR description
-   ✅ Rebase branch to HEAD
-   ✅ Squash changes into one signed, single commit

### DIFF
--- a/modules/cluster-api/graph/schema.resolvers.go
+++ b/modules/cluster-api/graph/schema.resolvers.go
@@ -306,10 +306,10 @@ func (r *subscriptionResolver) LogRecordsFollow(ctx context.Context, kubeContext
 
 		// Handle errors
 		if stream.Err() != nil {
-			// Log the error on the server
-			zlog.Error().Err(stream.Err()).Msg("Error during log stream")
+			// Log the error on the server, including caller info as requested.
+			zlog.Error().Err(stream.Err()).Caller().Send()
 
-			// If the client is still connected, let them know the stream ended due to an error
+			// If the client connection is still open, let it know that there was an upstream error.
 			if ctx.Err() == nil {
 				transport.AddSubscriptionError(ctx, gqlerrors.ErrInternalServerError)
 			}
--- a/modules/dashboard/graph/schema.resolvers.go
+++ b/modules/dashboard/graph/schema.resolvers.go
@@ -1040,10 +1040,10 @@ func (r *subscriptionResolver) LogRecordsFollow(ctx context.Context, kubeContext
 
 		// Handle errors
 		if stream.Err() != nil {
-			// Log the error on the server
-			zlog.Error().Err(stream.Err()).Msg("Error during log stream")
+			// Log the error on the server, including caller info as requested.
+			zlog.Error().Err(stream.Err()).Caller().Send()
 
-			// If the client is still connected, let them know the stream ended due to an error
+			// If the client connection is still open, let it know that there was an upstream error.
 			if ctx.Err() == nil {
 				transport.AddSubscriptionError(ctx, gqlerrors.ErrInternalServerError)
 			}


rxinui (dismissed):
/lgtm

modules/cluster-api/graph/schema.resolvers.go
```diff
@@ -306,7 +305,9 @@ func (r *subscriptionResolver) LogRecordsFollow(ctx context.Context, kubeContext
 
 		// Handle errors
 		if stream.Err() != nil {
-			transport.AddSubscriptionError(ctx, gqlerrors.ErrInternalServerError)
```
amorey, line 309: If the client connection is still open we should let it know that there was an upstream error: 
```go
if stream.Err() != nil {
  zlog.Error().Err(stream.Err()).Caller().Send()
  if ctx.Err() == nil {
    transport.AddSubscriptionError(ctx, gqlerrors.ErrInternalServerError)
  }
}
```

modules/cluster-api/graph/schema.resolvers.go
```diff
@@ -306,7 +305,9 @@ func (r *subscriptionResolver) LogRecordsFollow(ctx context.Context, kubeContext
...
+			// Just log the error on the server. The channel will be closed,
+			// and the client will know the subscription has ended.
+			zlog.Ctx(ctx).Error().Err(stream.Err()).Msg("Error during log stream")
```
amorey: I don't have much experience using `.Ctx()` but in our case we don't have a sub-logger registered in the context and ChatGPT says that in that case zerolog will return a no-op logger.

amorey (changes requested):

modules/cluster-api/graph/schema.resolvers.go
ramsyana: Thanks for the catch! I've pushed the fix. The error handling now logs the error and notifies the client as you suggested.

modules/cluster-api/graph/schema.resolvers.go
ramsyana: Thanks for the heads-up. The fix I just pushed avoids this issue by using the global zlog logger instead.

amorey (approved):

amorey:
Looks good! Thanks!

PR #451, opened 2025-06-11
Dashboard dual port support (http and https)

armadi1809:
🐋 Implement dashboard dual support (http and https)

Fixes # 388

## Summary

Add dual support for http and https in dashboard

## Changes

- Connect to both HTTP and HTTPS servers. 
- Config modifications to handle enabling/disabling ports individually
- Add self signed cert to dev environment manifest

## Submitter checklist

- [v] Add the correct emoji to the PR title
- [v] Link the issue number, if any, to *Fixes #*
- [v] Add summary and explain changes in the PR description
- [v] Rebase branch to HEAD
- [v] Squash changes into one signed, single commit [^1]


amorey (changes requested):
Can you disable your yaml auto-formatter and submit again? It's difficult to focus on what's actually changing with formatting changes mixed in.

hack/tilt/kubetail.yaml
```diff
@@ -45,9 +45,92 @@ data:
...
           hide-health-checks: true
       tls:
-        enabled: false
-        cert-file:
-        key-file:
+        enabled: true
+        cert-file: /etc/kubetail/tls/tls.crt
+        key-file: /etc/kubetail/tls/tls.key
+  # TLS certificate and key generated using - openssl req -x509 -newkey rsa:4096 -keyout tls.key -out tls.crt -days 365 -nodes -subj "/CN=localhost"
+  tls.crt: |
```
amorey, line 52: Let's split this out into a `Secret` of type tls (https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets)

amorey: Let's split this out into a `Secret` of type tls (https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets)

modules/shared/config/config.go
```diff
@@ -58,7 +58,10 @@ type Config struct {
 
 	// Dashboard options
 	Dashboard struct {
-		Addr               string   `validate:"omitempty,hostname_port"`
```
amorey, line 61: Let's change the config to have this structure:
```yaml
dashboard:
  http:
    enabled: true
    address: "0.0.0.0"
    port: 8080
  https:
    enabled: true
    address: "0.0.0.0"
    port: 8443
    tls:
      cert-file: "/etc/ssl/certs/server.crt"   # path to PEM-formatted certificate
      key-file:  "/etc/ssl/private/server.key" # path to PEM-formatted private key
```

modules/dashboard/cmd/main.go
```diff
@@ -145,15 +177,17 @@ func main() {
 			defer cancel()
 
 			var wg sync.WaitGroup
-			wg.Add(2)
+			wg.Add(len(servers) + 1) // +1 for app shutdown
```
amorey: How about calling `wg.Add()` before each goroutine is registered so we don't have to pre-determine the number here?

modules/dashboard/cmd/main.go
```diff
@@ -176,7 +210,10 @@ func main() {
 	flagset := cmd.Flags()
 	flagset.SortFlags = false
 	flagset.StringVarP(&cli.Config, "config", "c", "", "Path to configuration file (e.g. \"/etc/kubetail/dashboard.yaml\")")
-	flagset.StringP("addr", "a", ":8080", "Host address to bind to")
```
amorey, line 179: Let's remove all of the addr-related flags altogether. You can already set these values easily with a config file or by using the "param" flag.

hack/tilt/kubetail.yaml
```diff
@@ -44,10 +51,97 @@ data:
...
-        enabled: false
-        cert-file:
-        key-file:
+---
+apiVersion: v1
+kind: Secret
+metadata:
+name: secret-tls
+type: kubernetes.io/tls
+stringData::
```
amorey: If you want base64 encoded then use `data` instead (`stringData` is not base64 encoded). Either is ok but I guess `data` is preferable in this context just because it will make the file shorter.

hack/tilt/kubetail.yaml
```diff
@@ -237,10 +331,22 @@ spec:
           periodSeconds: 5
           failureThreshold: 3
         resources: {}
+      volumeMounts:
```
amorey: This needs to be moved one level deeper and merged with the existing `volumeMounts` (at L311)

hack/tilt/kubetail.yaml
```diff
@@ -237,10 +331,22 @@ spec:
...
+         readOnly: true
+       - name: tls
+         mountPath: /etc/ssl/private/server.key
+         subPath: tls.key
+         readOnly: true
       volumes:
       - name: config
         configMap:
           name: kubetail-dashboard
+      -name: tls
```
amorey: Need space in `-name:`

modules/shared/config/config.go
```diff
@@ -70,6 +69,25 @@ type Config struct {
 			ClusterAPIEnabled bool `mapstructure:"cluster-api-enabled"`
 		}
 
+		HTTP struct {
```
amorey, line 72: You don't need to prefix the mapstructure tag with the parent name of the struct (viper will handle that automatically). As written, you would have reference these variables in a config file is like this:
```yaml
dashboard:
  http:
    http-enabled: true
    http-port: 9000
```

Also, you don't need to include a `mapstructure` tag for simple strings so this is fine:
```go
HTTP struct {
  Enabled bool 
  Address string `validate:"omitempty,hostname"`
  Port uint `validate:"omitempty,port"`
}
```

modules/shared/config/config.go
```diff
@@ -237,7 +243,14 @@ func DefaultConfig() *Config {
 	cfg := &Config{}
 
 	cfg.AllowedNamespaces = []string{}
-	cfg.Dashboard.Addr = ":8080"
+	cfg.Dashboard.HTTP.Address = ""
+	cfg.Dashboard.HTTP.Port = 8080
+	cfg.Dashboard.HTTP.Enabled = true
+	cfg.Dashboard.HTTPS.Address = ""
+	cfg.Dashboard.HTTPS.Port = 8443
+	cfg.Dashboard.HTTPS.Enabled = true
```
amorey: Enabling HTTPS by default with an empty cert will throw errors when the main loop look tries to set up tls.

modules/dashboard/cmd/main.go
```diff
@@ -176,7 +212,8 @@ func main() {
...
+	flagset.Bool("http-enabled", true, "Enable HTTP server")
```
amorey: Let's remove these flags. HTTP and HTTPS can be configured with a config file or using the `param` CLI flag if necessary.

modules/dashboard/cmd/main.go
```diff
@@ -145,15 +178,18 @@ func main() {
 			defer cancel()
 
 			var wg sync.WaitGroup
-			wg.Add(2)
+			wg.Add(1) // +1 for app shutdown
```
amorey: Instead of declaring it here, just call `wg.Add()` before starting the app shutdown goroutine

modules/dashboard/cmd/main.go
```diff
@@ -65,6 +66,10 @@ func main() {
 			// Init viper
 			v := viper.New()
 			v.BindPFlag("dashboard.addr", cmd.Flags().Lookup("addr"))
+			v.BindPFlag("dashboard.http-addr", cmd.Flags().Lookup("http-addr"))
```
amorey: As per previous comment, let's remove the `http*` flags.

hack/tilt/kubetail.yaml
```diff
@@ -9,7 +9,14 @@ metadata:
 data:
   config.yaml: |
     dashboard:
-      addr: :8080
+      http-addr: 0.0.0.0
```
amorey: These values should be nested:
```yaml
dashboard:
  http:
    enabled: true
    address: 0.0.0.0
    port: 8080
  https:
    enabled: true
    address: 0.0.0.0
    port: 8443
    cert-file: /etc/ssl/certs/server.crt
    key-file: /etc/ssl/private/server.key
```

hack/tilt/kubetail.yaml
armadi1809: Should we also expose the https port here (L307)

hack/tilt/kubetail.yaml
```diff
@@ -44,10 +53,17 @@ data:
...
-        cert-file:
-        key-file:
+---
+apiVersion: v1
+kind: Secret
+metadata:
+name: secret-tls
+type: kubernetes.io/tls
+data:
+  tls.crt: |
```
amorey: The trailing spaces are throwing an error. Also, you don't need "|" for single-line entries.

hack/tilt/kubetail.yaml
```diff
@@ -44,10 +53,17 @@ data:
...
           enabled: true
           hide-health-checks: true
-      tls:
-        enabled: false
-        cert-file:
-        key-file:
+---
+apiVersion: v1
+kind: Secret
+metadata:
```
amorey: `name` must be nested under `metadata`. Also, `namespace` is missing. Use this:
```yaml
metadata:
  name: kubetail-dashboard-tls
  namespace: kubetail-system
```

hack/tilt/kubetail.yaml
```diff
@@ -237,10 +261,13 @@ spec:
           periodSeconds: 5
           failureThreshold: 3
         resources: {}
-      volumes:
-      - name: config
-        configMap:
-          name: kubetail-dashboard
+        volumes:
```
amorey: This indentation of the `volumes:` block needs to be pulled up one level.

hack/tilt/kubetail.yaml
```diff
@@ -9,7 +9,16 @@ metadata:
...
-      addr: :8080
+      http:
+        addr: 0.0.0.0
+        port: 8080
+        enabled: true
+      https:
+        addr: 0.0.0.0
+        port: 8443
+        enabled: true
+        cert-file: /etc/ssl/certs/server.crt
```
amorey: File args need to be nested under `tls`.

amorey:
It'd be great to merge this PR and the last remaining problems are just yaml configuration issues. Do you think you can fix those issues soon? If you're busy let me know and I can help.

amorey (approved):

Issue #417, opened 2025-06-09
CI: refactor build workflow


rxinui:
Currently, the CI lacks of best practices as every builds are enumerated per arch.

## TODO

Leverage github actions matrix strategy to simplify the build workflow for different arch.

See https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/ci.yml

It would be clever to use https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow rather than explicitly all jobs.

Take this opportunity to bring other refactoring within the workflow.

NickSolante:
Hey are you able to assign this to me?

amorey:
@NickSolante just assigned it to you

amorey:
Fixed in https://github.com/kubetail-org/kubetail/pull/477 (not sure why it didn't get closed automatically)

PR #459, opened 2025-06-12
:recycle: Split CI into 2 files to improve readability and reusability

NickSolante:
Fixes # [417](https://github.com/kubetail-org/kubetail/issues/417)

## Summary

This PR consolidates the GitHub Actions CI workflow to improve reliability and maintainability by replacing separate Docker build jobs with a unified matrix strategy and adding timeout protection to prevent hanging builds.

## Changes

- Consolidated Docker builds: Replaced separate docker-builds-amd64 and docker-builds-arm64 jobs with a single docker-builds job using matrix strategy
- Improved matrix organization: Each Docker image/architecture combination gets its own cache scope for better build performance
- Preserved existing functionality: All builds, tests, and CLI compilation remain unchanged

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -1,0 +1,88 @@
+name: 'Setup Environment'
+description: 'Sets up Node.js, PNPM, Go, and Rust conditionally for building Kubetail'
+
+inputs:
+  go-version:
+    description: 'Go version to install'
+    required: false
+    default: '1.24.3'
+  node-version:
+    description: 'Node.js version to install'
+    required: false
+    default: '22'
+  pnpm-version:
+    description: 'PNPM version to install'
+    required: false
+    default: '10'
+  setup-rust:
+    description: 'Whether to set up Rust'
+    required: false
+    default: 'false'
+  rust-toolchain:
+    description: 'Rust toolchain to install'
+    required: false
+    default: '1.87.0'
+  setup-protoc:
+    description: 'Whether to set up Protoc'
+    required: false
+    default: 'false'
+  setup-go:
+    description: 'Whether to set up Go'
+    required: false
+    default: 'false'
+  setup-node:
+    description: 'Whether to set up Node.js and PNPM'
+    required: false
+    default: 'false'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Protoc
+      if: inputs.setup-protoc == 'true'
+      uses: arduino/setup-protoc@v3
+
+    - name: Set up Rust
+      if: inputs.setup-rust == 'true'
+      uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: ${{ inputs.rust-toolchain }}
+        components: rustfmt, clippy
+
+    - name: Set up Rust cache
+      if: inputs.setup-rust == 'true'
+      uses: Swatinem/rust-cache@v2
+      with:
+        workspaces: "./crates/rgkl"
+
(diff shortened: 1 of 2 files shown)


.github/actions/setup-environment/action.yml
```diff
@@ -0,0 +1,72 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Protoc
+      if: inputs.setup-protoc == 'true'
+      uses: arduino/setup-protoc@v3
+
+    - name: Set up Rust
+      if: inputs.setup-rust == 'true'
+      uses: actions-rs/toolchain@v1
```
amorey: `actions-rs/toolchain` was archived in 2023

.github/workflows/ci.yml
```diff
@@ -9,473 +9,148 @@ on:
...
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
-      - uses: arduino/setup-protoc@v3
-      - name: Run builder
+      - uses: ./.github/actions/setup-environment
+        with:
+          setup-rust: 'true'
+          setup-protoc: 'true'
+          setup-go: 'false'
```
amorey: Removing arguments with default 'false' would make the action more readable

.github/actions/setup-environment/action.yml
```diff
@@ -0,0 +1,72 @@
...
+      with:
+        workspaces: "./crates/rgkl"
+
+    - name: Set up PNPM
+      uses: pnpm/action-setup@v4
+      with:
+        version: ${{ inputs.pnpm-version }}
+        run_install: false
+
+    - name: Set up Node.js
```
amorey, line 64: Did you consider making the node setup optional? Node isn't required for the rust or go checks.

amorey (changes requested):

NickSolante:
> There are some big improvements over the current set up here. Thanks!
> 
> I don't have any experience with splitting the action into two files so I'll defer to @rxinui for any comments there.
> 
> One overall question - previously the lint/vet/test steps were being done in parallel but now they're being done in series. I can see advantages of doing them in series but one disadvantage is that developers won't be made aware of all failures at once. Why did you decide to implement the checks in series?

Yeah fair, I see your point. I was thinking more about maximising the use of single runners for each flow (i.e., crates, modules, dashboard-ui). This makes maintenance possibly easier as single runners would fail sequentially, and adding new checks or modifying the flow would be easier. It's also nice that if linting fails, quality issues need to be addressed first - more expensive jobs like test suites don't get run unnecessarily. The series approach does provide a fail-fast feedback loop. 
That said, I understand the value of seeing all failures at once. Happy to keep that part as is - would you prefer to explore a hybrid approach, or should we stick with the current parallel setup?

amorey:
Awesome - thanks for making those changes. Can you rebase and sign your commit? Otherwise looks good to me.

amorey (approved):

amorey:
Thanks! This is a nice upgrade to our CI workflow. As a follow-on what do you think about using a 2-D matrix for the docker builds?

NickSolante:
> Thanks! This is a nice upgrade to our CI workflow. As a follow-on what do you think about using a 2-D matrix for the docker builds?

Yea initially I didn't think about it, but could definitely DRY up some of the workflow and make things even faster, was also looking at the modules-vet and test could split that out since that workflow handles all the go modules in a single job we could parallelise it to increase speed and visibility.

PR #477, opened 2025-06-15
♻️ 2D Docker Matrix and Modular Parallelisation 

NickSolante:
Fixes # [417](https://github.com/kubetail-org/kubetail/issues/417)

## Summary
This PR enhances the CI workflow by converting Docker builds to a 2D matrix structure and restoring per-module parallelization for Go modules to improve build visibility, maintainability, and performance.

## Changes

- Docker builds refactor: Converted from explicit include list to clean 2D matrix (build × runner) with 8 total combinations
- Modules parallelisation: Split modules-lint, modules-vet, and modules-test back to per-module matrix jobs for better failure isolation and speed
- Enhanced Docker features: Added debug builds, build-args, metadata generation, artifact uploads, and advanced caching to docker builds
- Improved build visibility: Each module and Docker variant now runs independently, making it easier to identify specific failures
- Maintained functionality: All existing builds, tests, and CLI compilation preserved with same coverage

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,36 +59,45 @@ jobs:
 
   modules-lint:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        module: [ cli, cluster-agent, cluster-api, dashboard, shared ]
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
         with:
           setup-go: 'true'
-      - name: Run linter
-        working-directory: ./modules
+      - name: Run linter for ${{ matrix.module }}
+        working-directory: ./modules/${{ matrix.module }}
         run: test -z "$(gofmt -l .)"
 
   modules-vet:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        module: [ cli, cluster-agent, cluster-api, dashboard, shared ]
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
         with:
           setup-go: 'true'
-      - name: Run code vetter
-        working-directory: ./modules
-        run: go vet github.com/kubetail-org/kubetail/modules/...
+      - name: Vet ${{ matrix.module }}
+        working-directory: ./modules/${{ matrix.module }}
+        run: go vet ./...
 
   modules-test:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        module: [ cli, cluster-agent, cluster-api, dashboard, shared ]
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
         with:
           setup-go: 'true'
-      - name: Run tests
-        working-directory: ./modules
-        run: go test -race github.com/kubetail-org/kubetail/modules/...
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -142,53 +151,89 @@ jobs:
...
+          file: ${{ matrix.build.dockerfile }}
+          platforms: ${{ matrix.platform }}
+          target: ${{ matrix.build.target }}
           push: false
-          tags: kubetail-${{ matrix.image }}:${{ matrix.tag-suffix }}
-          cache-from: type=gha,scope=${{ matrix.cache-scope }}
-          cache-to: type=gha,mode=max,scope=${{ matrix.cache-scope }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: ${{ matrix.build.build-args }}
```
amorey: Is `build-args` being used? If not, let's remove it.

.github/workflows/ci.yml
```diff
@@ -142,53 +151,89 @@ jobs:
...
+          cache-from: |
+            type=gha,scope=${{ matrix.build.image }}-${{ matrix.build.target }}-${{ matrix.arch }}
+            type=local,src=/tmp/.buildx-cache
+          cache-to: |
+            type=gha,mode=max,scope=${{ matrix.build.image }}-${{ matrix.build.target }}-${{ matrix.arch }}
+            type=local,dest=/tmp/.buildx-cache-new,mode=max
+          outputs: type=docker,dest=/tmp/${{ matrix.build.image }}-${{ matrix.build.target }}-${{ matrix.arch }}.tar
+
+      - name: Move cache
```
amorey: What is this step for?

.github/workflows/ci.yml
```diff
@@ -142,53 +151,89 @@ jobs:
...
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+
+      - name: Upload image artifact
```
amorey: Will the artifacts be used for anything later?

NickSolante: The artifacts could be used for manual testing specific builds or running a test suite against the specific artifact, with some more thought probably best to remove this

.github/workflows/ci.yml
NickSolante: We don't need this anymore since we removed the type=local cache and only use type=gha now. The move step was for Docker buildx's local cache limitation where it can't read and write to the same local path. With just GHA cache, buildx handles everything automatically.

NickSolante:
Ah, you're absolutely right - that was an oversight on my part! I included both `type=gha` and `type=local` without really thinking through whether we actually needed the local caching.
The `type=local` cache doesn't really help us in this matrix setup since each job runs on separate runners with their own filesystems. The local cache starts empty every time, so we were just adding complexity with the cache shuffling for no real benefit. The `type=gha` cache is doing all the actual work - that's what persists between builds and gives us the speed improvements. Should have just stuck with that from the start instead of cargo-culting the dual cache approach

amorey (approved):
Looks great!

PR #479, opened 2025-06-16
🐋 test: add transport_test

### LINKED ISSUE #386: Write unittests for custom Kubernetes bearer token transport
In order to attach end-users' bearer tokens to internal Kubernetes requests we are using this custom `http.RoundTripper`: https://github.com/kubetail-org/kubetail/blob/main/modules/shared/k8shelpers/transport.go

It's a core part of the code that is used everywhere but currently it doesn't have any unit tests. We should add some.

atlomak:
Fixes #386 

## Summary
This PR adds test for custom transport

## Changes
Added new test which checks if custom authorization transport adds authorization header from request's context.

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/k8shelpers/transport_test.go
+++ b/modules/shared/k8shelpers/transport_test.go
@@ -30,7 +30,6 @@ func TestBearerTokenRoundTripper_headerSet(t *testing.T) {
 	ctx := context.WithValue(req.Context(), K8STokenCtxKey, TEST_TOKEN)
 	req = req.WithContext(ctx)
 
-	c.Do(req)
+	_, err = c.Do(req)
 	assert.Nil(t, err)
-
 }


modules/shared/k8shelpers/transport_test.go
```diff
@@ -0,0 +1,36 @@
...
+	c := &http.Client{Transport: b}
+
+	req, err := http.NewRequest("GET", testserver.URL, nil)
+	assert.Nil(t, err)
+
+	ctx := context.WithValue(req.Context(), K8STokenCtxKey, TEST_TOKEN)
+	req = req.WithContext(ctx)
+
+	c.Do(req)
+	assert.Nil(t, err)
```
amorey, line 34: `err` is already being checked at L28

modules/shared/k8shelpers/transport_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var TEST_TOKEN = "eyJhbGciOiJSUzI1NiIsImtpZCI6IjEyMzQ1NiJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwic3ViIjoic3lzdGVtOnNvbWUtbmFtZXNwYWNlOnRlc3QtdXNlciIsImF1ZCI6Imt1YmVybmV0ZXMiLCJleHAiOjQ3Mjg5MjQ4MDAsImlhdCI6MTY1MDkzMjAwMH0.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
+
+func TestBearerTokenRoundTripper_headerSet(t *testing.T) {
+	testserver := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		header, ok := r.Header["Authorization"]
```
amorey, line 16: The test doesn't wait for this function to be called before exiting. To ensure that the function gets executed during the test you can use a `sync.WaitGroup`.

amorey (changes requested):

amorey:
Thanks for the PR! I added a couple of comments. When you commit again make sure you sign the commit (`-s`).

modules/shared/k8shelpers/transport_test.go
atlomak: Yes, I overlooked missing `err` assignment to value returned from `c.Do(req)` Thanks

modules/shared/k8shelpers/transport_test.go
atlomak: Isn't `c.Do(req)` blocking, so the handler function has to return something before ending the test?

amorey: Sorry, you're right.

amorey (approved):

amorey:
Thanks for the PR! 🥳

PR #497, opened 2025-06-19
🐋 Rust agent workspace - DO NOT MERGE 

freexploit:
This is an initial commit to start working on the rust cloud agent.


amorey:
Awesome start to the project! Can I suggest that we call it "cluster-agent" instead of "agent" to be consistent with our other cluster-api/ cluster-agent naming?

freexploit:
Oh Yes, I'm doing the change now

freexploit:
Yes, please

build/package/Dockerfile.cluster-agent
```diff
@@ -20,16 +20,14 @@ WORKDIR /work
 RUN apt-get update && apt-get install -yq protobuf-compiler
 
 # Install Rust dependencies (for cache)
-COPY crates/rgkl/Cargo.toml ./crates/rgkl/Cargo.toml
-COPY crates/rgkl/Cargo.lock ./crates/rgkl/Cargo.lock
-COPY crates/rgkl/src/main.rs ./crates/rgkl/src/main.rs
+COPY crates/ ./crates
```
amorey, line 23: If we copy `Cargo.toml` and `Cargo.lock` files (and I think you need `main.rs` too) before running `cargo fetch` then Docker will cache the results of the step. Otherwise we'll re-refetch dependencies everytime the source code changes even if the dependencies are the same.

crates/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
+resolver = "3"
```
amorey, line 2: Just curious - what does "resolver" refer to?

freexploit: That is Cargo resolver version. 3 is the latest on 2024 edition. [https://doc.rust-lang.org/cargo/reference/resolver.html](https://doc.rust-lang.org/cargo/reference/resolver.html) It has to be defined on workspaces because crates may or may not share dependencies.

build/package/Dockerfile.cluster-agent
freexploit: This is because now I copy all the workspace instead of just rgkl. This affects because now that is building target path is now on workspace root instead of crate directory. Also now the protobuf files are on types crate so I need to have the workspace configuration for this. I think We can do better job caching all that, but to be able to keep rgkl backward compatible I decided to copy the whole create directory.

gikaragia (approved):
LGTM!

build/package/Dockerfile.cluster-agent
amorey: Copying the entire workspace makes sense but if you do that after copying the dependency files and running `cargo fetch` then we can take advantage of docker caching. Anyways, we can implement this optimization later when we're ready to push to production.

PR #509, opened 2025-06-21
🐋 Add dual port support to cluster-api server (HTTP + HTTPS)

### LINKED ISSUE #389: Add dual port support to Cluster API server (http & https)
We would like to modify the Cluster API server (https://github.com/kubetail-org/kubetail/tree/main/modules/cluster-api) so that it listens on both http and https ports by default. The purpose of this is to allow for dual http+https setups which is helpful in debugging and other scenarios.

In order to do this we need to make the following modifications:
- Enable connections to both ports in server (https://github.com/kubetail-org/kubetail/blob/main/modules/dashboard/cmd/main.go)
- Modify config to handle enabling/disabling ports individually (https://github.com/kubetail-org/kubetail/blob/main/modules/cluster-api/cmd/main.go)
- Add hard-coded, self-signed SSL certificate to dev environment manifest (https://github.com/kubetail-org/kubetail/blob/main/hack/tilt/kubetail.yaml)

tzabbi:
Fixes #389 

## Summary

Add http and https support for cluster-api-server.

## Changes

- adding the ability to use http and/or https port for cluster api server
- both ports are enabled by default
- update README.md of dashboard and cluster-api go module

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -212,6 +212,7 @@ k8s_resource(
     'kubetail-cluster-api:clusterrolebinding',
     'kubetail-cluster-api:role',
     'kubetail-cluster-api:rolebinding',
+    'kubetail-cluster-api-tls:secret',
   ],
   resource_deps=['kubetail-shared'],
 )
--- a/hack/config.yaml
+++ b/hack/config.yaml
@@ -12,11 +12,9 @@ allowed-namespaces: []
 ## dashboard ##
 #
 dashboard:
-
   ## http ##
   #
   http:
-
     ## enabled ##
     #
     # Default value: true
@@ -27,7 +25,7 @@ dashboard:
     #
     # Default value: __empty
     #
-    address: 
+    address:
 
     ## port ##
     #
@@ -38,7 +36,6 @@ dashboard:
   ## https ##
   #
   https:
-
     ## enabled ##
     #
     # Default value: false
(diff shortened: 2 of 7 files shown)


tzabbi:
Yeah, I can do this. Can you tell me the reason for this request? I didn't specify any special formatting, it is just the default rules, which I highly recommend to use.

amorey:
> Yeah, I can do this. Can you tell me the reason for this request? I didn't specify any special formatting, it is just the default rules, which I highly recommend to use.

It's hard to see the config changes when they're mixed in with formatting changes.

What formatter are you using? I'm not opposed to autoformatting yaml but it would be useful to specify the rules somewhere so they're consistent for everyone. Also if we make rule changes we should format all the yaml files in the repo at the same time otherwise the files will be inconsistent with each other.

amorey:
@tzabbi Interested in setting up a yaml linter? https://github.com/kubetail-org/kubetail/issues/510

tzabbi:
> It's hard to see the config changes when they're mixed in with formatting changes.

I totally agree with this. The reason I think it is worth do have these changes is that there are some best practices like no trailing whitespace at the end of line and so on.

> What formatter are you using?

I use the https://github.com/redhat-developer/yaml-language-server in both of my IDE's nvim and zed.

tzabbi:
> @tzabbi Interested in setting up a yaml linter? #510

YES!!! I love to implement linter and other qa tools like trivy, ...

PR #512, opened 2025-06-23
Draft: 🐋 Add yaml-lint github action and add .editorconfig

### LINKED ISSUE #510: Implement repo-wide yaml linting rules
Currently we don't have any enforced yaml linting rules so we've had a few PRs that made yaml formatting changes to individual files due to the contributor's autoformatter. It would be useful to configure repo-wide yaml linting rules so that all contributors can use the same rules. We should also fix yaml linting issues to conform to new rules.

Todo:
[ ] Add yaml linting config to repo that can be used by autoformatters
[ ] Fix lint issues and run auto formatter on existing yaml files

tzabbi:
Fixes #510 

## Summary

Add yaml-lint github action job and .editorconfig

## Changes

- to have a standardized YAML code style, we implement YAML lint GitHub action
- to have some unified code rules through all files, we add a .editorconfig
## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+ident_style = space
+ident_size = 2
+
+[yaml]
+quote_type = double
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,8 +4,6 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf-8
-ident_style = space
-ident_size = 2
 
 [yaml]
 quote_type = double
--- a/.github/workflows/yaml-lint.yml
+++ b/.github/workflows/yaml-lint.yml
@@ -2,29 +2,20 @@
 name: YAML Lint
 
 on:
-  push:
-    paths:
-      - "**.yml"
-      - "**.yaml"
   pull_request:
     paths:
       - "**.yml"
       - "**.yaml"
 
 jobs:
-  yamllint:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.x"
-
       - name: Install yamllint
         run: pip install yamllint
 
       - name: Run yamllint
-        run: yamllint .
+        run: yamllint --config-file .yamllint.yml .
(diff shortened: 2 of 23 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+ident_style = space
```
amorey: type on `indent_style` and `indent_size`

.github/workflows/yaml-lint.yml
```diff
@@ -0,0 +1,30 @@
+---
+name: YAML Lint
+
+on:
+  push:
```
amorey: I think we can remove the check on "push". We require pull requests for all merges into `main` so the check on "pull_request" should be sufficient.

.github/workflows/yaml-lint.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: Install yamllint
```
amorey, line 17: Does it make sense to use a one-step yamllint action? I don't have any personal experience with them but here are some options:
https://github.com/marketplace/actions/yaml-lint
https://github.com/marketplace/actions/yamllint-github-action

amorey (changes requested):
Let's ignore `pnpm-lock.yaml` and other generated yaml files because otherwise we'll be in constant contention with the code generators

amorey (commented):
The current .editorconfig settings will contend with the gofmt tool that uses tabs instead of spaces. Considering that this is a multi-language monorepo does it make more sense to configure each language formatter separately?

tzabbi:
@amorey do you have some requirements for rust, go, ... files in regard of code syntax? I will add them to the .editorconfig. Same for the yaml linter, which files will be autogenerated?

amorey:
@tzabbi Currently we're language-specific linters and vetters with their default setups (see [ci.yml](https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/ci.yml)).

.editorconfig
tzabbi: Removed it.

.github/workflows/yaml-lint.yml
tzabbi: Done!

.github/workflows/yaml-lint.yml
tzabbi: I just used the example from the yaml lint documentation:   https://yamllint.readthedocs.io/en/latest/integration.html#integration-with-github-actions

Issue #493, opened 2025-06-19
Update Development Environment for new Rust-based Cluster Agent


freexploit:
Modify the development environment to support the new Rust-based Cluster Agent.

- [ ] Dockerfile
- [ ] Tiltfile
- [ ] Github workflow
- [ ] Helm Chart

gikaragia:
Going through the rest of the tasks to see what's left @amorey.

I took a quick look at the helm charts, and I think that we don't need to update them for the new cluster agent as the app has the same outside behavior. Do you agree?

amorey:
Yeah, in theory we shouldn't need to change the templates.

gikaragia:
With #625 this is now done

PR #518, opened 2025-06-27
Allow the new cluster agent to be deployed in the cluster

gikaragia:
- 🐋 New feature

## Summary

Enables the deployment of the new cluster agent in the local cluster. This is just to enable running tests on the cluster while developing. There is more cleanup work needed for the deployment scripts, but I left this work to be done as part of #493

## Changes
- Fixes the IP address to be exposed to other containers
- Copies the new cluster agent when using KUBETAIL_DEV_RUST_LOCAL
- I removed the health check for specific services as the k8s scripts check for the basic health service only. We can revise this decision if we ever need to do any per-service health-checking/recovery.

## Submitter checklist

- [ ] Rebase branch to HEAD

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -36,30 +36,13 @@ docker_build_with_restart(
 
 # kubetail-cluster-agent
 
-local_resource(
-  'kubetail-cluster-agent-compile',
-  '''
-  cd modules
-
-  # --- Build Go binary   ---
-  export CGO_ENABLED=0
-  export GOOS=linux
-  go build -o ../.tilt/cluster-agent ./cluster-agent/cmd/main.go
-  ''',
-  deps=[
-    './modules/cluster-agent',
-    './modules/shared'
-  ]
-)
 
 build_rust_locally = os.getenv("KUBETAIL_DEV_RUST_LOCAL", default='false').lower() == 'true'
 if build_rust_locally:
   local_resource(
-    "kubetail-rgkl-compile",
+    "kubetail-cluster-agent-compile",
     '''
     set -eu
-
-    cd crates/rgkl
 
     # --- Determine target architecture ---
     arch=$(uname -m)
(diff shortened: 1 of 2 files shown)


Tiltfile
```diff
@@ -70,15 +53,19 @@ if build_rust_locally:
...
 
     # --- Copy to .tilt directory ---
-    out_dir="../../.tilt"
+    out_dir="../.tilt"
     mkdir -p "$out_dir"
     cp "target/${target}/debug/rgkl" "$out_dir"
+    cp "target/${target}/debug/cluster_agent" "$out_dir/cluster-agent"
     ''',
     deps=[
+      "./crates/cluster_agent/src",
```
amorey, line 68: A few questions/comments:
* Do we need to add the cluster agent's `Cargo.toml` and `Cargo.lock` too?
* Do we need to add `crates/types` as a dependency too?
* Does it make sense to just use the entire `crates` directory as a dependency?

Tiltfile
```diff
@@ -101,13 +88,29 @@ if build_rust_locally:
     ]
   )
 else:
+  local_resource(
```
amorey, line 91: We should be able to get rid of this entire block now

Tiltfile
gikaragia: Yes to all 🙂

Tiltfile
gikaragia: Yeah, I was thinking about that too. I wasn't sure, though, if we want to keep the option of having a development environment that doesn't need Rust dependencies as we do with the KUBETAIL_DEV_RUST_LOCAL variable now so I left the decision for the other issue. I could do this now though.

So WDYT? Should I drop this option together with the Dockerfile.kubetail-cluster-agent file? Or should I keep it as an option?

amorey: I think a modified form of the existing behavior in `main` should work:
* `KUBETAIL_DEV_RUST_LOCAL=true` - use `Dockerfile.kubetail-cluster-agent-local` for the base image and compile the `cluster-agent` binary locally
* `KUBETAIL_DEV_RUST_LOCAL=false` - use `Dockerfile.kubetail-cluster-agent` for the base image and let docker compile the Rust executable inside the image

The code in this PR handles `KUBETAIL_DEV_RUST_LOCAL=true` correctly but now that we don't need the Go binary anymore we can just get rid of this block for both cases `true/false`:
```
local_resource(
  'kubetail-cluster-agent-compile',
  '''
  cd modules

  # --- Build Go binary   ---
  export CGO_ENABLED=0
  export GOOS=linux
  go build -o ../.tilt/cluster-agent ./cluster-agent/cmd/main.go
  ''',
  deps=[
    './modules/cluster-agent',
    './modules/shared'
  ]
)
```

amorey: Also - we can delete the `modules/cluster-agent` directory entirely and see where things break.

PR #522, opened 2025-06-30
Implement LogRecords in rust cluster agent

### LINKED ISSUE #492: Implement LogRecordService in Rust
Port the `LogRecordService` functionality from Go to Rust.

gikaragia:
Fixes #492

## Summary

Implements the LogRecord service in Rust cluster agent. As the PR was starting to get a bit big, I think that it would be a good idea to handle some of the tasks in other issues. Namely:
- Implement TLS
- Add logging (currently I am doing prints)
- Allow the cluster agent to read the configuration from yaml files.
- Look into using jemallocator again as we did previously (which is coming from [this comment](https://github.com/BurntSushi/ripgrep/blob/master/crates/core/main.rs#L19)). The problem with using it currently was that I couldn't build the cluster agent for alpine as I was an error that was saying:  "background_threads_runtime_support not supported for x86_64-unknown-linux-musl"

## Changes
- I tried to mostly keep existing code the same to make sure that I don't introduce bugs.
- I removed the rgkl binary as it makes sense to only have it as a library now
- The changes are mostly updating rgkl to work with tonic.

## Testing
- Ensure that the tests are passing (`cargo test`)
- Try the two local builds (with flag UBETAIL_DEV_RUST_LOCAL=true or not)

### Testing the service
As the integration with the UI does not work yet, follow the steps below to test the gRPC services (although there might be a better way to do it):
- Install [grpcurl](https://github.com/fullstorydev/grpcurl)
- Start the dev environment `tilt up`
- Get the loggen pod name, container id, and the cluster agent pod name with:
```
kubectl get pods --all-namespaces \
  --field-selector spec.nodeName=minikube-m03 \
  -o json | jq -r '
    .items[] | 
    . as $pod |
    $pod.status.containerStatuses[]? | 
    "\($pod.metadata.namespace)/\($pod.metadata.name) \(.containerID)"
'
```
- Map the cluster agent port to localhost with `kubectl -n kubetail-system port-forward <cluster agent pod name> 50051:50051`
- Finally, run a gRPC request with
```
grpcurl -plaintext -d '{
  "namespace":  "default",
  "podName":  "<loggen pod name>",
  "containerName":  "loggen",
  "containerId":  "containerd://<loggen container id>",
  "followFrom":  "END"
}
' localhost:50051 cluster_agent.LogRecordsService/StreamForward
```

## Submitter checklist

- [ ] Rebase branch to HEAD

Signed-off-by: Giannis K <ikaragi23@yahoo.gr>

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -96,6 +96,7 @@ else:
       './proto',
     ],
     ignore=[
+      './crates/target',
       './crates/*/target'
     ],
     live_update=[
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,10 +164,11 @@ jobs:
             image: cluster-api
             target: final
             dockerfile: build/package/Dockerfile.cluster-api
-          - name: cluster-agent-final
-            image: cluster-agent
-            target: final
-            dockerfile: build/package/Dockerfile.cluster-agent
+          #   TODO: Update Docker images with the rust client
+          # - name: cluster-agent-final
+          #   image: cluster-agent
+          #   target: final
+          #   dockerfile: build/package/Dockerfile.cluster-agent
         runner: [ ubuntu-24.04, ubuntu-24.04-arm ]
         include:
           - runner: ubuntu-24.04


crates/cluster_agent/src/main.rs
```diff
@@ -10,24 +13,51 @@ use logrecords::LogRecords;
...
 
     let reflection_service = tonic_reflection::server::Builder::configure()
         .register_encoded_file_descriptor_set(FILE_DESCRIPTOR_SET)
         .build_v1()?;
 
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::WARN)
+        .init();
+
+    let (term_tx, _term_rx) = broadcast::channel(5);
```
amorey: Do we need to buffer the broadcast channel?

Tiltfile
```diff
@@ -70,20 +53,23 @@ if build_rust_locally:
...
     ''',
     deps=[
-      "./crates/rgkl/src",
-      "./crates/rgkl/Cargo.toml",
-      "./crates/rgkl/Cargo.lock",
+      "./crates",
       "./proto",
     ],
+    ignore=[
+      './crates/target'
```
amorey: Should we also add the `target` subdirectories here (`'./crates/*/target'`)?

Tiltfile
```diff
@@ -107,7 +93,7 @@ else:
     context='.',
     entrypoint="/cluster-agent/cluster-agent -c /etc/kubetail/config.yaml",
     only=[
-      './crates/rgkl',
```
amorey, line 110: `'./.tilt/cluster-agent'` should be sufficient in this list (no need for `'./proto'` either) because the container only needs the final executable to run.

crates/cluster_agent/src/main.rs
gikaragia: Probably not, I picked a value that I considered safe at the time. I'll test with 1 and update

amorey: Sounds good, mostly I'm just curious because in Go I don't use channel buffering unless it's necessary because it can hide bugs.

gikaragia: Fixed in d4e85412150181d71235a2011ce0815039389a63

Tiltfile
gikaragia: Fixed in 506ad253d96c8c0fe699e3784f01ba5f432ae872

Tiltfile
gikaragia: The comment is correct but for the other flow. This one needs the files to build them in the container.

Fixed in 506ad253d96c8c0fe699e3784f01ba5f432ae872

Tiltfile
```diff
@@ -107,12 +92,11 @@ else:
...
+      './crates',
       './proto',
-      './.tilt/cluster-agent'
     ],
     ignore=[
-      './crates/rgkl/target'
+      './crates/*/target'
```
amorey, line 100: I think we need to add './crates/target' here too

amorey (changes requested):

Tiltfile
gikaragia: Fixed in 43b54f78ed3d6f6e9a23349384de5f02eab63ffa

amorey (approved):

PR #546, opened 2025-07-15
🐋 Add mTLS support to Cluster Agent

amorey:
## Summary

This PR adds mTLS support to the Cluster Agent gRPC server and TLS support to the Cluster API gRPC client. This PR also enables tls by default in the dev environment using self-signed certificates. After this PR I will revert the dual http port support in the Dashboard and Cluster Agent in a separate PR. Then everything will be in place to enable adding the Cluster API as a Kubernetes API extension (`APIService`).

## Changes

* Add self-signed TLS certs for development to repo
* Add self-signed TLS certs to tilt resources
* Modify config to handle mTLS gRPC connections in Cluster Agent and Cluster API
* Add support for mTLS connections to Cluster Agent gRPC server
* Add support for TLS connections to Cluster API gRPC client
* Add unit tests for Cluster Agent gRPC server
* Switch kubernetes built-in gRPC readinessProbe for exec call with grpc-health-probe

## Submitter checklist

- [x] Rebase branch to HEAD


hack/tilt/tls/README.md
```diff
@@ -0,0 +1,57 @@
+# Dev TLS
+
+## CA commands
+
+1. Generate the CA private key:
```
rxinui, line 5: nice!

hack/tilt/tls/ca.crt
```diff
@@ -0,0 +1,35 @@
...
+O9mmOcgZrmOUhJ5Jw1CgBIhovnaKqqXjsugjUoWvHWXbZY5FMHRwy1n7s5u14Byd
+lU3YmdgK2/0+00TNls9IYOQl1UM9SykG6mGjCQJTqE+srUVX3+mOaomWuA4/0Yba
+0UwAU8YmFuvWvccMSCB83a3xgUc/DLUwhOseb6pw7GAMSjCto5qrhICqinCMWu6h
+HJGFhOMKevGCvFB+b3mJJ1K8Be9PsNQIuvKNBHINqdfOuFHbBEe74NHVy9AajehO
+grSZh74YTIaS2CPBPtTaBvlDBkBSXvHMd0kbtYScIrjX+SCE7TAKvYNVx0wK3y13
+MCfo9B3kP5VGvbVNJm0FGwk3jxLRMPKYhbAS2GPikD/cDH1fDTiLoCpzse3Qjs9A
+Y0LlSCag68hVtCfF4OMIRK9MzLVlvnhdxuhz7ZsGdvzbxrcJVdsRyfXoT+H/1Br/
+OsVKY/sj6mCNLShP8LkVLaZCGm/RXyEERL2BJF1akMRrY0xN/fGFFuzl2aptiQkM
+qbuCE0NkRGuqxk7ZlODpx/MbA4ZMdr1324kdtQclJEZ5Z/v8lki9vg==
+-----END CERTIFICATE-----
```
rxinui, line 35: even if it is for tilt, why isnt it ignore?

modules/cluster-agent/README.md
```diff
@@ -22,16 +22,18 @@ The Kubetail Cluster Agent executable supports the following command line config
...
+| allowed-namespaces           | []string | If populated, restricts namespace access | []       | stable |
+| cluser-agent.addr            | string   | Host address to bind to                  | ":50051" | stable |
+| cluser-agent.logging.enabled | bool     | Enable logging                           | true     | stable |
+| cluser-agent.logging.level   | string   | Log level                                | "info"   | stable |
+| cluser-agent.logging.format  | string   | Log format (json, pretty)                | "json"   | stable |
+| cluser-agent.tls.enabled     | bool     | Enable TLS endpoint termination          | false    | stable |
+| cluser-agent.tls.cert-file   | string   | Path to cert file                        | ""       | stable |
+| cluser-agent.tls.key-file    | string   | Path to key file                         | ""       | stable |
+| cluser-agent.tls.ca-file     | string   | Path to client CA bundle file            | ""       | alpha  |
+| cluser-agent.tls.client-auth | string   | Controls client cert authentication      | "none"   | alpha  |
```
rxinui, line 36: nice to add that

modules/cluster-agent/internal/server/server_test.go
```diff
@@ -0,0 +1,238 @@
...
+	assert.NoError(t, err)
+	assert.NotNil(t, server)
+	assert.IsType(t, &grpc.Server{}, server)
+}
+
+func TestNewServerTLSConfiguration(t *testing.T) {
+	// Create temporary directory for test certificates
+	tempDir, err := os.MkdirTemp("", "server_test_*")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDir)
```
rxinui: we should do that instead testing facilitates it for us

Use `t.TempDir()` to delegate and create the tempdir by the testing lib.

amorey: Thanks! I wasn't aware of `t.TemptDir()`

hack/tilt/tls/ca.crt
amorey: I just updated the code to use `secret_create_generic()` and `secret_create_tls()` from tilt so now the secrets are defined at runtime from the files in `hack/tilt/tls`.

hack/tilt/kubetail.yaml
```diff
@@ -525,28 +508,49 @@ spec:
...
+              readOnly: true
             - name: varlibdockercontainers
               mountPath: /var/lib/docker/containers
               readOnly: true
           readinessProbe:
-            grpc:
-              port: 50051
+            exec:
+              command:
+                - /bin/grpc_health_probe
```
gikaragia, line 518: Hey @amorey as I'll look into implementing this for the new cluster agent, can you clarify if I understood this change correctly?

So previously we were using the native k8s gRPC health checking but now we moved to using this tool in order to enable mTLS? So the idea is that k8s will execute this command in the cluster agent's pod which in turn will check if the local gRPC service is enabled?

amorey: Yes, exactly. The native k8s gRPC health probe doesn't support mTLS so we're using a third-party probe instead ([grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe)). The probe is added to the docker image at build time (https://github.com/kubetail-org/kubetail/pull/546/files#diff-3fd2f96cd92c9290b318f53fad596a5a5d11ec6c88e40378e3306216755772a8R82) and invoked using the `exec` feature.

amorey: For background - the reason we need mTLS support is because it's required as part of the Kubernetes API extension layer. When we add the Cluster API as a Kubernetes API extension, all client requests to the Cluster API will get routed through the Kubernetes API proxy server first which requires an mTLS connection. The Kubernetes API proxy will perform authentication and add impersonation headers to the request before forwarding it to the Cluster API. The Cluster API which will in turn forward the impersonation headers to the Cluster Agent so securing the connection between the Cluster API and the Cluster Agent with mTLS will ensure that all the requests in the chain are from authorized clients.

PR #547, opened 2025-07-16
🎣 Switch to using `SyncMap` in desktop connection manager to fix slow startup times

amorey:
## Summary

The goal of this PR is to fix intermittent slow startup times in the Kubetail Dashboard when running in a desktop environment. The cause is lock contention in `DesktopConnectionManager` across clusters due to the the singleton `sync.Mutex` used by instances of the struct. To fix this issue, this PR adds threadsafe `LoadOrCompute()` and `LoadOrComputeWithContext()` methods to the `SyncMap` implementation and uses them to replace the global `sync.Mutex` in `DesktopConnectionManager`.

Note: `LoadOrCompute()` is essentially a thread-safe memoization method so it seems like there should already be something in the standard library or a third-party implementation we can use instead. When I looked around I found the [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) library but it doesn't support generics and isn't context-aware so it's not ideal. There are other versions with those features (e.g. [janos/singleflight](https://github.com/janos/singleflight)) but none are popular and the implementation isn't difficult so I decided to implement it here. We should definitely revisit this if the upstream singleflight library gets updated or there's a better way to do this that I missed.

## Changes

* Adds new `LoadOrCompute()` and `LoadOrComputeWithContext()` methods to `SyncMap`
* Refactors `DesktopConnectionManager` and replace use of global `sync.Mutex` with `SyncMap` instances
* Fixes previous mistake where `DesktopConnectionManager.WaitUntilReady()` was ignoring the context argument

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/k8shelpers/connection-manager.go
+++ b/modules/shared/k8shelpers/connection-manager.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/utils/ptr"
 
 	"github.com/kubetail-org/kubetail/modules/shared/config"
+	"github.com/kubetail-org/kubetail/modules/shared/util"
 )
 
 // Represents shared informer factory cache key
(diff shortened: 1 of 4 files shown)


amorey:
Thanks for the review.

The underlying issue is that I was seeing slow startup times locally and I think it was due to lock contention in the `WaitUntilReady()` method which initializes a kubernetes clientset before releasing the lock. The slow code is in the REST config initializer which invokes exec plugins when present (like in my local setup): https://github.com/kubetail-org/kubetail/blob/4335bb08b428d8af76011e2499ea371e520fc7c9/modules/shared/k8shelpers/connection-manager.go#L283-L288. Maybe there was a nice way to prevent the issue but keep the global lock but like you said, SyncMap will make things easier to reason about and maintain.

PR #556, opened 2025-07-24
✨ Use tracing for logging

### LINKED ISSUE #524: Use tracing for logging
We should use tracing for logging. We currently do println! everywhere.

hakymulla:
Fixes #524 

## Summary

This PR introduces the use of  tracing crate for logging

## Changes

 Tracing crate for logging instead of println!

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/logmetadata.rs
+++ b/crates/cluster_agent/src/logmetadata.rs
@@ -134,16 +134,16 @@ impl LogMetadataService for LogMetadataImpl {
 
 #[cfg(test)]
 mod test {
+    use crate::logmetadata::LogMetadataImpl;
     use std::io::Write;
     use tempfile::{Builder, NamedTempFile};
     use tokio::sync::broadcast;
     use tokio_util::task::TaskTracker;
     use tonic::Request;
+    use tracing_test::traced_test;
     use types::cluster_agent::{
         LogMetadataListRequest, log_metadata_service_server::LogMetadataService,
     };
-    use tracing_test::traced_test;
-    use crate::logmetadata::LogMetadataImpl;
 
     fn create_test_file(name: &str, num_bytes: usize) -> NamedTempFile {
         let mut test_file = Builder::new()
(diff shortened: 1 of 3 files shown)


crates/cluster_agent/src/logmetadata.rs
```diff
@@ -195,9 +198,11 @@ mod test {
         assert_eq!("pod-name", log_metadata.spec.as_ref().unwrap().pod_name);
         assert_eq!("container-name", log_metadata.spec.unwrap().container_name);
         assert_eq!(4, log_metadata.file_info.unwrap().size);
+        assert!(logs_contain("Filename could not be parsed:"));
```
gikaragia: So the test here uses a temp directory, which probably has other unrelated files, and this line will get logged.

However, if the temp directory is empty, then this test case will fail. I think that the tests that use TermReader are enough and we shouldn't bother doing any log checks here.

gikaragia (changes requested):
Overall it looks great @hakymulla, just a small change.

If you are looking for any other issues to pick up, let me know!

crates/cluster_agent/src/logmetadata.rs
```diff
@@ -250,5 +255,6 @@ mod test {
         let log_metadata = result.items.pop().unwrap();
 
         assert_eq!("secondnamespace", log_metadata.spec.unwrap().namespace);
+        assert!(logs_contain("Filename could not be parsed:"));
```
gikaragia: Same here

gikaragia (approved):
LGTM! 

Before merging this, there is no sign-off on your most recent commit, and two sign-offs in the previous one:
```
    Use tracing for logging

    new: use tracing for logging and test

    * new: replace println! with tracing macros
    * test: add #[traced_test] for log assertions in tests

    Signed-off-by: Kazeem Hakeem <hakymulla@gmail.com>
    Signed-off-by: Hakeem Kazeem <hakymulla@gmail.com>
```

cc @amorey

PR #558, opened 2025-07-25
🎣 Fix concurrent memoization issue in syncmap

amorey:
## Summary

This PR fixes an issue preventing `LoadOrCompute()` and `LoadOrComputeWithContext()` callback functions in the `SyncMap` implementation from running concurrently even for different map keys. Previously the `SyncMap` implementation used a per-instance mutex to implement wait-until-ready behavior which was preventing callbacks for different map keys from running concurrently. The solution is to implement per-key mutex locks but rather than adding this to `SyncMap` itself we decided to simplify the struct and return it to its original intension which is a type-aware implementation of `sync.Map`. Instead, we moved `LoadOrCompute()` and `LoadOrComputeWithContext()` into a new per-key thread-safe sister struct called `SyncGroup` that properly implements the desired memoization behavior.

This PR also does some light refactoring.

## Changes

* Remove `LoadOrCompute()` and `LoadOrComputeWithContext()` from `SyncMap`
* Create new struct `SyncGroup` with new methods that uses `SyncMap` under-the-hood to implement per-key mutex locks
* Removes unnecessary loop var copy statements from `shared/k8shelpers/connection-manager.go`
* Removes unnecessary logging statement from DesktopConnectionManager cache warmer

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/k8shelpers/connection-manager.go
+++ b/modules/shared/k8shelpers/connection-manager.go
@@ -21,7 +21,6 @@ import (
 	"sync"
 	"time"
 
-	zlog "github.com/rs/zerolog/log"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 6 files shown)


amorey:
Thanks!

PR #563, opened 2025-07-28
🎣 Issue warning on browser open error instead of fatal exit

### LINKED ISSUE #531: Browser open error running `kubetail serve` on CentOS
When I ran `kubetail serve` on CentOS I got the following error:
```sh
$ kubetail serve
2025-07-07T18:19:16.904575014-04:00 INF Started kubetail dashboard on http://localhost:7500/
2025-07-07T18:19:16.904670265-04:00 FTL error="exec: \"xdg-open,x-www-browser,www-browser\": executable file not found in $PATH"
```

It's possible that migrating from [github.com/pkg/browser](https://pkg.go.dev/github.com/pkg/browser) to a well maintained fork might fix the problem 🤞

ASaidOguz:
## Summary

Previously, the Kubetail CLI `serve` sub-command would issue a fatal exit if it was unable to open a new browser tab. This PR modifies the behavior and issues a warning instead of exiting the command loop.

## Changes

* Replaced `zlog.Fatal()` with `zlog.Warn()` in `modules/cli/cmd/serve.go`

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/cli/cmd/serve.go
+++ b/modules/cli/cmd/serve.go
@@ -170,7 +170,7 @@ var serveCmd = &cobra.Command{
 		if !skipOpen {
 			err = browser.OpenURL(fmt.Sprintf("http://%s:%d/", host, port))
 			if err != nil {
-				zlog.Fatal().Err(err).Send()
+				zlog.Warn().Err(err).Msg("Unable to open browser automatically. Please open the dashboard URL manually.")
 			}
 		}
 


modules/cli/cmd/serve.go
```diff
@@ -170,7 +170,11 @@ var serveCmd = &cobra.Command{
 		if !skipOpen {
 			err = browser.OpenURL(fmt.Sprintf("http://%s:%d/", host, port))
 			if err != nil {
-				zlog.Fatal().Err(err).Send()
+				// If browser fails to open, log the error but continue
+				// this is not critical for the server operation
+				// we can still access the dashboard via the URL
+				//zlog.Fatal().Err(err).Send()
```
amorey: We don't need to leave this line in

modules/cli/cmd/serve.go
```diff
@@ -170,7 +170,11 @@ var serveCmd = &cobra.Command{
 		if !skipOpen {
 			err = browser.OpenURL(fmt.Sprintf("http://%s:%d/", host, port))
 			if err != nil {
-				zlog.Fatal().Err(err).Send()
+				// If browser fails to open, log the error but continue
```
amorey: I don't think we need these comments because the `zlog.Warn()` line is self-explanatory.

amorey (changes requested):

amorey:
Thanks for the PR. I made a couple of small comments on the code change.

How did you debug the issue on CentOS? Did you test against a full version of CentOS running with a GUI? The version I tested with had a browser available.

amorey:
Also, to fulfill the DCO requirement you can roll back your commit and re-commit it with your signature:

```console
git reset --soft HEAD~1
git commit -a -s -m "fix: log warning instead of fatal exit on browser open error in kubetail cli serve sub-command"
git push origin --force
```

amorey:
> Tested with the shell script i created so exposed the ports to my wsl linux environment and i could reached the browser from it.Used centos stream 10 image. We may choose to act like this in specificaly for this error as warning using regex of the error and crash app for any other errors.But need to further study for browser packages specific errors . Its completly your call .

It makes sense that a CentOS container image wouldn't have a browser but I'm also getting an error when I try it in a full CentOS installation with a browser. For example, here is CentOS running inside a UTM virtual machine on my mac:
https://www.loom.com/share/a1ff5aee31974877914d8270db8aa5f1?from_recorder=1&focus_title=1

I installed docker and minikube, then ran `kubtail serve`.

amorey:
@ASaidOguz Looks like the PR description got mangled. Let me know when you want me to review.

amorey:
> why would it crash the application even if the go code doesnt lead it to ?
did u properly build the binary ser ?

The video uses the latest version of `kubetail`, not your branch. I just wanted to show that there's an error when you run `kubetail serve` even if there's a browser available.

PR #564, opened 2025-07-29
🎣 ci: fix #551 by authenticating git and enabling Cargo CLI fetch

### LINKED ISSUE #551: Fix "API rate limit exceeded" errors in GitHub actions
Currently we're getting errors like this when we run our GitHub workflows:

```
Error: Error: API rate limit exceeded for 68.220.61.179. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

The error is happening somewhat frequently (~30% of runs). It would be really nice to fix this so we don't have to re-run our workflows frequently.

SANTHOSH-MAMIDISETTI:
Fixes #551 – API rate limit errors in GitHub Actions

## Summary
This PR resolves intermittent `API rate limit exceeded` errors (~30% of runs) by:
1. Authenticating all `git` fetches using `GITHUB_TOKEN`
2. Enabling `CARGO_NET_GIT_FETCH_WITH_CLI=true` to avoid libgit2 API calls
3. Applying the env at both job-level and composite action level for extra reliability

## Changes
- `.github/actions/setup-environment/action.yml`
  - Added git authentication step
  - Added `CARGO_NET_GIT_FETCH_WITH_CLI=true` for Rust steps
- `.github/workflows/ci.yml`
  - Added job-level `CARGO_NET_GIT_FETCH_WITH_CLI` for all Rust jobs

## Testing
- ✅ Draft PR will validate on real GitHub runners

## Expected Outcome
- CI runs without hitting `API rate limit exceeded` errors
- Less flakiness and fewer reruns required

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -38,6 +38,11 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Authenticate git for GitHub
+      shell: bash
+      run: |
+        git config --global url."https://${{ secrets.GITHUB_TOKEN }}:@github.com/".insteadOf "https://github.com/"
+
     - name: Set up Protoc
       if: inputs.setup-protoc == 'true'
       uses: arduino/setup-protoc@v3
@@ -48,6 +53,8 @@ runs:
       with:
         toolchain: ${{ inputs.rust-toolchain }}
         components: rustfmt, clippy
+      env:
+        CARGO_NET_GIT_FETCH_WITH_CLI: "true" # Optional: Avoid API rate limits for libgit2 by using CLI
 
     - name: Set up Rust cache
       if: inputs.setup-rust == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ on:
 jobs:
   crates-build:
     runs-on: ubuntu-24.04
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
@@ -23,6 +25,8 @@ jobs:
 
   crates-lint:
     runs-on: ubuntu-24.04
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
@@ -35,6 +39,8 @@ jobs:
 
   crates-vet:
     runs-on: ubuntu-24.04
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-environment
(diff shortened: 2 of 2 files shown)


SANTHOSH-MAMIDISETTI:
Please review when convenient  @amorey — this should eliminate the API rate-limit flakiness in CI.

Thanks!

SANTHOSH-MAMIDISETTI:
✅ Update: I've removed all `CARGO_NET_GIT_FETCH_WITH_CLI` env var usage from both the CI workflow and the `setup-environment` composite action as suggested, 
This should keep things clean while still resolving the API rate limit issues we were seeing earlier.

Let me know if there's anything else needed  @amorey .

Thanks again!

SANTHOSH-MAMIDISETTI:
Sure thing , will look into it, thanks for the info!

amorey:
@tharpta Welcome to the party! We're not having problems with the `olegtarasov/get-tag` action. The problem we're trying to solve is that sometimes one of our Rust actions hits a rate limit and we think it's because cargo is making unauthenticated calls to GitHub when installing dependencies.

amorey:
@SANTHOSH-MAMIDISETTI It looks like you updated your branch by merging `main` into your branch. In this case I don't think it matters because we will do a squash merge back into main so we'll preserve linear history in the `main` branch but in general you should use a "rebase" to preserve linear history.

SANTHOSH-MAMIDISETTI:
> @SANTHOSH-MAMIDISETTI It looks like you updated your branch by merging `main` into your branch. In this case I don't think it matters because we will do a squash merge back into main so we'll preserve linear history in the `main` branch but in general you should use a "rebase" to preserve linear history.

Thanks for the heads-up! You're right — a rebase would’ve been the better option to keep the history clean. I’ll make sure to use git rebase instead of merging main into my branch going forward.

.github/actions/setup-environment/action.yml
```diff
@@ -38,6 +38,11 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Authenticate git for GitHub
```
amorey:
```suggestion
    - name: Set up git authentication for GitHub requests in Rust environment
       if: inputs.setup-rust == 'true' && inputs.GITHUB_TOKEN != ''
       shell: bash
      run: |
        git config --global url."https://${{ inputs.GITHUB_TOKEN }}:@github.com/".insteadOf "https://github.com/"
```

amorey (changes requested):

.github/actions/setup-environment/action.yml
```diff
@@ -38,6 +38,15 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Set up git authentication for GitHub requests in Rust environment
+       if: inputs.setup-rust == 'true' && inputs.github-token != ''
+       shell: bash
+      run: |
+        git config --global url."https://${{ inputs.GITHUB_TOKEN }}:@github.com/".insteadOf "https://github.com/"
+      shell: bash
```
amorey, line 47: We need to remove L46-48

amorey (changes requested):
In addition to removing L46-48 we also need to add `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` everywhere `setup-rust` is used (e.g. https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/ci.yml#L21)

amorey:
@SANTHOSH-MAMIDISETTI Would you be able to rebase and make those changes (remove extra lines at L46-48, add github token to `setup-rust` actions)? Let me know if you're busy and would like some help finishing up the PR.

PR #568, opened 2025-08-02
🎣 Bugfixes and improvements to RGKL LogFileRegexMatcher

### LINKED ISSUE #565: Search code panics when it encounters multiple identical lines
Using [test.log](https://github.com/user-attachments/files/21533100/test.log):

```console
$ cargo run -- stream-forward ~/scratch/search-issue/test2.log --grep yyy
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.05s
     Running `target/debug/rgkl stream-forward test.log --grep yyy`
$ cargo run -- stream-forward ~/scratch/search-issue/test2.log --grep zzz
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.05s
     Running `target/debug/rgkl stream-forward /Users/andres/scratch/search-issue/test2.log --grep zzz`
{"timestamp":"2025-07-31T12:06:00.001936471Z","message":"zzz"}

thread 'main' panicked at src/index.crates.io-1949cf8c6b5b557f/grep-printer-0.2.2/src/json.rs:631:31:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

amorey:
Fixes #565 

## Summary

This PR fixes a bug causing the LogFileRegexMatcher to panic when it encounters multiple identical lines. The problem was that our implementation of the find_at() method required by the `Matcher` trait assumed the `haystack` only consisted of one log line when actually it can consist of multiple lines as matches are accumulated. I believe that this behavior exists so that you can write matchers that operate on several lines simultanously but I'm not 100% sure (TODO: dig into grep crate to really understand it). In anycase, I modified the code to ignore multi-line haystacks and also made some improvements to the matching logic to simplify the code. It works now 

## Changes

* Return no matches when haystack consists of multiple log lines
* Added unit tests to test message extraction code for CRI and Docker formatted log lines

## Submitter checklist

- [x] Rebase branch to HEAD


amorey:
FYI @gikaragia @freexploit - a user is waiting on this fix so I'm going to merge the PR but it would be useful to do a deeper dive into this code at some point. I still don't entirely understand the behavior of the `find_at()` function in the `grep::Matcher` (e.g. how can we prevent the haystack from growing in size as matches are found) so it would be nice to get more eyes on it.

PR #570, opened 2025-08-03
🐋 Removes token-based csrf protection in favor of `Sec-Fetch-Site` header

amorey:
## Summary

This PR removes all token-based CSRF protection code from the frontend and the backend and replaces it with `Sec-Fetch-Site` header-based CSRF protection which is easier to support. In our case, data requests are handled through GraphQL APIs which aren't vulnerable to CSRF attacks so the only reason we have for CSRF protection is to prevent cross-site websocket hijacking (CSWSH). CSWSH protection with tokens is cumbersome because you can't send the token as part of the initial upgrade request so you have to send it separately after the websocket connection is initiated. In addition, the library we were using to support token-based CSRF protection on the server ([gorilla/csrf](github.com/gorilla/csrf)) was just updated to fix a vulnerability ([[CVE-2025-24358](https://github.com/advisories/GHSA-rq77-p4h8-4crw)] in a way that doesn't fix a vulnerability we have and makes the library unusable so we would be unable to use it going forward. On the other hand, the `Sec-Fetch-Site` header is supported by all modern browsers which is what we're targeting anyway so using that to implement CSWSH protection is much simpler.

## Changes

* Removes all token-based CSRF protection code from the frontend (dashboard-ui) and the backend (modules/dashboard, modules/cluster-api, modules/shared)

## Submitter checklist

- [x] Rebase branch to HEAD


modules/cluster-api/internal/app/app.go
```diff
@@ -108,44 +105,11 @@ func NewApp(cfg *config.Config) (*App, error) {
...
-				c.JSON(http.StatusOK, gin.H{"value": csrf.Token(c.Request)})
-			})
-		}
-
 		// authentication middleware
 		dynamicRoutes.Use(authenticationMiddleware)
 
 		// GraphQL endpoint
-		app.graphqlServer = graph.NewServer(app.cm, app.grpcDispatcher, cfg.AllowedNamespaces, csrfProtect)
+		app.graphqlServer = graph.NewServer(cfg, app.cm, app.grpcDispatcher, cfg.AllowedNamespaces)
```
rxinui, line 112: that looks cleaner indeed

modules/cluster-api/graph/server_test.go
```diff
@@ -20,67 +20,73 @@ import (
...
+		wantStatus     int
+	}{
+		{
+			"csrf disabled, non-browser client",
+			false,
+			http.Header{},
+			http.StatusSwitchingProtocols,
+		},
+		{
+			"csrf disabled, same-origin request",
```
rxinui, line 43: oh i see

modules/dashboard/graph/server.go
```diff
@@ -91,35 +83,27 @@ func NewServer(config *config.Config, cm k8shelpers.ConnectionManager, csrfProte
...
-				// If the header is absent, we fall back to the CSRF token validation
-				// in the InitFunc. This supports older browsers or non-browser clients.
+				// If empty, request is from non-browser or legacy browser
 				if secFetchSite == "" {
 					return true
 				}
 
-				// For modern browsers that send the header, enforce strict same-site policies.
-				// This is the primary defense against Cross-Site WebSocket Hijacking (CSWSH).
+				// Check the Sec-Fetch-Site header as our primary defense against
```
rxinui, line 98: ok

rxinui (approved):

PR #351, opened 2025-05-29
🐋 Add support for multiple config files to KubeConfigWatcher

amorey:
Fixes #344

## Summary

This PR adds support for multiple config files to KubeConfigWatcher and wraps the file system event handler with a debounce wrapper to avoid triggering multiple callbacks as part of a single config file modification write.

## Changes

- Modifies KubeConfigWatcher to provide support for multiple config files via the KUBECONFIG env var
- Adds a debounce method to KubeConfigWatcher fsnotify handler to deal with potential of multiple file events as part of a single write
- Adds unit test to KubeConfigWatcher to verify support for multiple config files
- Adds unit test to KubeConfigWatcher to verify support for file changes

## Submitter checklist

- [x] Add the correct emoji to the PR title
- [x] Link the issue number, if any, to *Fixes #*
- [x] Add summary and explain changes in the PR description
- [x] Rebase branch to HEAD
- [x] Squash changes into one signed, single commit [^1]

[^1]: See suggested [commit format](https://github.com/kubetail-org/.github/blob/main/pull-request-commit-format.md)

PR #572, opened 2025-08-04
🐋 Implement Watch of LogMetadataService

### LINKED ISSUE #491: Implement LogMetadataService in new Rust-based Cluster Agent
Port the `LogMetadataService` functionality from Go to Rust.

Related tech spec: https://github.com/orgs/kubetail-org/discussions/447

gikaragia:
Fixes #491

## Summary

Implements the watch service for LogMetadata. 

There are some limitations to this service that stem from how the filesystem works. From investigating online it seems that we should consider events coming from notify more like an event of "something has changed" as the events are generally not consistent. Some of these inconsistencies can be found in [notify's documentation](https://docs.rs/notify/latest/notify/).

Events are debounced by using notify's debouncer every 2 seconds for each file. As we are mapping many event types, to only three types, there is still some possibility of duplicate events currently.

There are some improvements that I will implement as part of a subsequent PR:
- Improve error handling
- Deduplicate events using an IndexSet.

## Testing instructions
- Ensure that the tests are passing (`cargo test`)

### Testing the service
- Install [grpcurl](https://github.com/fullstorydev/grpcurl)
- Start the dev environment `tilt up`
- Get the loggen pod name, container id, and the cluster agent pod name with:
```
kubectl get pods --all-namespaces \
  --field-selector spec.nodeName=minikube-m03 \
  -o json | jq -r '
    .items[] | 
    . as $pod |
    $pod.status.containerStatuses[]? | 
    "\($pod.metadata.namespace)/\($pod.metadata.name) \(.containerID)"
'
```
- Map the cluster agent port to localhost with `kubectl -n kubetail-system port-forward <cluster agent pod name> 50051:50051`
- Finally, run a gRPC request with
```
grpcurl -plaintext -d '{
  "namespaces":  ["default"]
}
' localhost:50051 cluster_agent.LogMetadataService/Watch
```

## Submitter checklist

- [ ] Rebase branch to HEAD

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -383,6 +383,7 @@ dependencies = [
  "prost-wkt-types",
  "regex",
  "rgkl",
+ "serial_test",
  "tempfile",
  "thiserror 2.0.12",
  "tokio",
@@ -683,6 +684,7 @@ checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
 dependencies = [
  "futures-channel",
  "futures-core",
+ "futures-executor",
  "futures-io",
  "futures-sink",
  "futures-task",
@@ -2579,6 +2581,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "scc"
+version = "2.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22b2d775fb28f245817589471dd49c5edf64237f4a19d10ce9a92ff4651a27f4"
+dependencies = [
+ "sdd",
+]
+
+[[package]]
 name = "schannel"
 version = "0.1.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


amorey:
Nice - I followed your instructions and got it working locally with grpcurl. I think the duplicate event issue is also a problem with the Go `fsnotify` package (https://github.com/kubetail-org/kubetail/pull/351). I'll dig into the code but it will take me some time to get up-to-speed.

gikaragia:
> Is there a way to use a ready signal instead of waiting 100ms in the tests?
Hmm, not sure how we can do that without polluting production code with test code. Any ideas?

> Should we replace .unwrap() before going live?
Good point! So the plan is to look in more detail at error handling in subsequent PRs. I started taking a look already and my plan for next PRs is:
- Define dependencies in the workspace level and upgrade to the latest version of prost, tonic, tokio, etc
- Define our own errors and look into detail all error paths
- Hande the deduplication that I mentioned in the description.

> Did you rebase on top of https://github.com/kubetail-org/kubetail/pull/556?
Yep!

amorey:
> Hmm, not sure how we can do that without polluting production code with test code. Any ideas?

I don't but I can look into it.

> Good point! So the plan is to look in more detail at error handling in subsequent PRs. I started taking a look already and my plan for next PRs is:
> 
>     * Define dependencies in the workspace level and upgrade to the latest version of prost, tonic, tokio, etc
> 
>     * Define our own errors and look into detail all error paths
> 
>     * Hande the deduplication that I mentioned in the description.

Sounds good!

PR #580, opened 2025-08-08
✨Upgrade fsnotify to v1.9.0 #487.

DavidReque:
Fixes #487 

## Summary
Upgrade fsnotify to v1.9.0 

## Changes

- **Upgrade**  
  - Upgraded `fsnotify` to `v1.9.0` in `shared` and `cluster-agent` (direct).  
  - Ran `go mod tidy` in `cli`, `cluster-api`, and `dashboard`.

## Cross-platform Tests

- **server_test.go**  
  - Use `assert.ErrorIs(err, os.ErrNotExist)` instead of string comparisons.

- **kube-config-watcher_test.go**  
  - Use `os.PathListSeparator` for `KUBECONFIG`.

- **helpers_test.go** and **logmetadata_test.go**  
  - Skip symlink-dependent tests on Windows.

## Watcher

- Unified to `event.Has()` for `Create` / `Write` / `Remove`.
- Ready to handle `Chmod` / `Rename` if macOS CI requires.

## CI

- Added new workflow `go-tests-macos.yml` running `go test ./...` per module on `macos-latest`.

## Submitter checklist

- [x] Add the correct emoji to the PR title
- [x] Link the issue number, if any, to *Fixes #*
- [x] Add summary and explain changes in the PR description
- [x] Rebase branch to HEAD
- [x] Squash changes into one signed, single commit [^1]

[^1]: See suggested [commit format](https://github.com/kubetail-org/.github/blob/main/pull-request-commit-format.md)

PR #610, opened 2025-08-21
🎣  Pin fsnotify to v1.8.0 to fix macos regression

amorey:
## Summary

This PR pins fsnotify to v1.8.0 to fix a macos regression introduced in v1.9.0 (https://github.com/fsnotify/fsnotify/issues/689). We had upgraded to v1.9.0 in a previous commit (https://github.com/kubetail-org/kubetail/pull/580) but it turned out we were only testing the affected code on linux so the regression went unnoticed for a while. Since the affected code is only in the Cluster Agent we could technically deprecate macos support but this will degrade the developer experience unless we also exclude the `module/cluster-agent` tests from running locally on macos as well. This isn't necessary for now but we can revisit the decision later.

## Changes

* Pin fsnotify to v1.8.0 using `replace` directive in go.mod files
* Remove macos-15 and macos-14 from CI workflow module test exclusion for cluster-api and cluster-agent

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,14 +101,6 @@ jobs:
         os: [ubuntu-24.04, ubuntu-24.04-arm, macos-15, macos-14, macos-13, windows-latest]
         module: [cli, cluster-agent, cluster-api, dashboard, shared]
         exclude:
-          - os: macos-15
-            module: cluster-api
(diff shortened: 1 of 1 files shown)


amorey:
@rxinui Merged small pr

PR #614, opened 2025-08-21
🐋Allow env variable substitution

### LINKED ISSUE #608: Allow environment configuration overrides
We should allow the cluster agent to be configured with environment variables.

The environment variables should start with CLUSTER-AGENT and should override the values written from the configuration files.

gikaragia:
Fixes #608

## Summary

Updates the configuration to be populated with env variables

## Changes
- File is now loaded by our code instead of config to grab the contents as a string
- We use subst crate to replace the env variables

## Testing instructions
- Add an env variable to `kubetail.yaml`, for example:
```yaml
      containers:
        - name: kubetail-cluster-agent
          image: kubetail-cluster-agent
          ports:
            - name: grpc
              protocol: TCP
              containerPort: 50051
          env:
            - name: NODE_NAME
              valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
+            - name: DEBUG_LEVEL
+              value: debug
```
- Add the variable for the cluster agent config file (both $DEBUG_LEVEL and ${DEBUG_LEVEL} should work):
```yaml
  labels:
    app.kubernetes.io/component: cluster-agent
data:
  config.yaml: |
    cluster-agent:
      addr: :50051
      logging:
        enabled: true
-       level: info
+       level: $DEBUG_LEVEL
        format: pretty
      tls:
        enabled: true
```
- Observe that the change in the configuration works

## Submitter checklist

- [ ] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/config.rs
+++ b/crates/cluster_agent/src/config.rs
@@ -1,13 +1,14 @@
 use std::{
     error::Error,
     io,
-    net::SocketAddr,
+    net::{AddrParseError, SocketAddr},
     path::{Path, PathBuf},
 };
 
-use config::builder::DefaultState;
+use config::{File, FileFormat, builder::DefaultState};
 use regex::Regex;
 use serde::Deserialize;
+use tokio::fs;
 
 #[derive(Debug)]
 pub struct Config {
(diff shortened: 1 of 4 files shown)


crates/cluster_agent/src/config.rs
```diff
@@ -114,4 +119,26 @@ impl Config {
             .set_default("cluster-agent.logging.format", "json")?
             .set_default("cluster-agent.tls.enabled", false)
     }
+
+    fn get_format(path: &Path) -> Result<FileFormat, Box<io::Error>> {
```
gikaragia, line 123: Ideally the transformation from file path to file format should be handled by the library to be able to support new formats seamlessly, but there isn't a public API for that unfortunately

PR #618, opened 2025-08-23
🐋Implement k8s authorization

### LINKED ISSUE #616: Implement k8s authorization
Currently, when we serve a request we don't verify that the client has the required k8s permissions.

We should add a step using k8s rust client to do this verification, similarly with the go client.

gikaragia:
Fixes #616

## Summary

Similarly with the old cluster agent, we now check that the pods making a request have the proper rights.

## Changes
- Adds a new struct, the authorizer which is responsible for communicating with the k8s to confirm the requests are authorized.

## Testing instructions
- Observe first that everything is still working
- Go to kubetail.yaml and remove the watch or list verb from the kubetail-cli cluster role:
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: kubetail-cli
  labels:
    app.kubernetes.io/component: cli
rules:
  - apiGroups: [""]
    resources: [nodes]
-    verbs: [get, list, watch]
+   verbs: [get, watch]
  - apiGroups: ["", apps, batch]
    resources:
      - cronjobs
      - daemonsets
      - deployments
      - jobs
      - pods
      - replicasets
      - statefulsets
    verbs: [get, list, watch]
  - apiGroups: [""]
    resources: [pods/log]
-   verbs: [list, watch]
+  verbs: [ watch]
```
- Observe that now the dashboard is broken

## Submitter checklist

- [ ] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/authorizer.rs
+++ b/crates/cluster_agent/src/authorizer.rs
@@ -1,0 +1,107 @@
+use k8s_openapi::api::authorization::v1::{
+    ResourceAttributes, SelfSubjectAccessReview, SelfSubjectAccessReviewSpec,
+};
+use kube::{Api, Client, Config, api::PostParams, config::AuthInfo};
+use tonic::{Status, metadata::MetadataMap};
+
+pub struct Authorizer {
+    k8s_config: Config,
+}
+
+/// Checks that the the k8s doing the request has proper rights to access the log files.
+#[cfg(not(test))]
+impl Authorizer {
+    /// Creates a new Authorizer, using the k8s authorization token to construct the proper
+    /// client set during authorization.
+    pub async fn new(request_metadata: &MetadataMap) -> Result<Self, Status> {
+        let token = request_metadata
+            .get("authorization")
+            .and_then(|token| token.to_str().ok())
+            .ok_or_else(|| {
+                Status::new(
+                    tonic::Code::Unauthenticated,
+                    "authentication token not found",
+                )
+            })?
+            .to_owned();
+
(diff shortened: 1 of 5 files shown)


crates/cluster_agent/src/authorizer.rs
```diff
@@ -0,0 +1,107 @@
...
+};
+use kube::{Api, Client, Config, api::PostParams, config::AuthInfo};
+use tonic::{Status, metadata::MetadataMap};
+
+pub struct Authorizer {
+    k8s_config: Config,
+}
+
+/// Checks that the the k8s doing the request has proper rights to access the log files.
+#[cfg(not(test))]
```
gikaragia, line 12: There are a couple of issues with this implementation.
- First, the ideal way to implement such functionality is with an interceptor (middleware). The problem is that tonic [does not support](https://github.com/hyperium/tonic/pull/910) async interceptors out of the box, and I would need to use a [custom crate](https://crates.io/crates/tonic-async-interceptor) for this functionality. I chose not to use an interceptor.
- Secondly, the ideal way to stub the authorizer in the tests would be to use dependency injection. However, this made the types very complicated, so I think that using #[cfg] directives is preferable

amorey: Makes sense. When we switch over to treating the Cluster API as a "Kubernetes API extension" then authentication will work with impersonation headers rather than tokens so I think doing things quick-and-simple like this for now sounds good. When we have a better idea of how auth will work long-term we can revisit it.

PR #623, opened 2025-08-29
📜 Add docs and address argument for cluster agent

### LINKED ISSUE #609: Create new cluster agent documentation

gikaragia:
Fixes #609

## Summary

Added documentation in the new cluster agent

## Changes
- I copied the docs from the existing cluster agent as it is the same functionality
- I added an addr argument to the cluster as it was stated in the docs
- I removed the documentation from rgkl as it is not meant to be used as cmd tool anymore

## Testing instructions
- In kubetail.yaml modify the entry for addr configuration for cluster agent to use another port.
- Observe that now cluster agent doesn't work.
- In the Tiltfile, add another argument to the entrypoint of cluster agent like: `-a :50051` or `--addr :50051` 
- Observe that the cluster agent works normally.

## Submitter checklist

- [ ] Rebase branch to HEAD

Signed-off-by: Giannis K <ikaragi23@yahoo.gr>


crates/rgkl/README.md
```diff
@@ -1,161 +0,0 @@
-# **rgkl - RipGrep for Kubernetes Logs**
```
gikaragia, line 1: I wasn't totally sure about deleting this. Any suggestions welcome.

amorey: It has some useful commands for development. Let's leave it for now and move/improve/remove it later.

gikaragia: The problem is that they don't work anymore, as rgkl isn't a command-line tool

gikaragia: We could potentially reintroduce the functionality, though, as the library is mostly the same as before.

amorey: Ahh I see. I could see re-introducing the functionality later but for now leaving it out and removing the docs sounds fine to me.

PR #626, opened 2025-08-31
✨Merge Rust cluster-agent implementation

gikaragia:
## Summary

Merges the new cluster agent. Information about the project can be found [here](https://github.com/orgs/kubetail-org/discussions/447)

## Changes

- Implements the cluster agent in Rust
- Makes rgkl a library and stops using it as a command.

## Testing instructions
Testing should ensure that no new discrepancies are introduced as part of this rewrite and there is a feature parity between the Go cluster agent and the new one.

To execute any tests:
- Check out this branch.
- Follow the instructions outlined [in the main repo](https://github.com/kubetail-org/kubetail#next-steps).
- Go through the UI and ensure everything works as it should.

Areas that should be tested:
- Main dashboard
- Searching for strings
- Tailing individual nodes
- Documentation

## Submitter checklist

- [x] Rebase branch to HEAD


gikaragia:
I had to implement some changes, mainly to fix the GitHub actions. These can be reviewed in commit 5401f432763058715023087b07906a79a10151cf

amorey:
Here's what I did to test things out locally (MBA M3):
- Ran tilt up (without `KUBETAIL_DEV_RUST_LOCAL`) [ok]
- Played with dashboard home page
  - Observed size and last event timestamps updating in real-time [ok]
  - Observed behavior for new pod creation [see below]
  - Performed page reload several times [ok]
  - Restarted the "kubetail-cluster-agent" resource in Tilt [see below]
- Played with dashboard logging console page
  - Observed default behavior of log streaming [ok]
  - Observed behavior when new pod is created [ok]
  - Jump to beginning [ok]
  - Jump to end [ok]
  - Relative time filter [ok]
  - Absolute time filter [ok]
  - Pause/play [ok]
  - Search [ok]
- Observed CPU + memory usage using `kubectl top pods -n kubetail-system` [ok]

Overall:
It's awesome! Memory usage has gone down from ~30-50MB to ~10MB and CPU is also down slightly from ~12m to ~10m. Also the executable is 7MB. 🔥

Notes:
- It would be nice to see an INFO logging message at startup to indicate that the server is working (current message is "Starting cluster-agent on :50051").
- Currently the agent is logging things such as client disconnect and watcher stop as INFO. I think this should be DEBUG otherwise logs will be very noisy by default.
- There's a bug where log metadata doesn't update in realtime for new pods. When I observed a new pod getting created I saw the initial size and initial event but subsequent changes didn't get updated in realtime.

gikaragia:
Thank you for the thorough testing! 

>  There's a bug where log metadata doesn't update in realtime for new pods. When I observed a new pod getting created I saw the initial size and initial event but subsequent changes didn't get updated in realtime.

How did you create the new pod so I can reproduce?

amorey:
> How did you create the new pod so I can reproduce?

New instances of `loggen` and `loggen-ansi` will get created automatically by the `chaoskube` agent every ~2 minutes but if you want to trigger it manually you can use `kubectl` do delete one of the pods in the `loggen-ansi` deployment (e.g. `kubectl delete pods/loggen-ansi-XXXX-YYYY`).

amorey:
> I think this was caused by the same issue, I tried it after the fix and it now recovers normally.

Nice

crates/cluster_agent/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+        overrides.push(("addr".to_owned(), address.to_owned()));
+    }
+
+    let config = Config::parse(config_path, overrides).await?;
+
+    Ok(config)
+}
+
+fn parse_overrides(param: &str) -> Result<(String, String), String> {
+    if let Some((name, value)) = param.split_once('=') {
```
amorey: The Go code uses ":" as a delimeter

crates/cluster_agent/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+    let (_, agent_health_service) = tonic_health::server::health_reporter();
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(FILE_DESCRIPTOR_SET)
+        .build_v1()?;
+    let (term_tx, _term_rx) = broadcast::channel(1);
+    let task_tracker = TaskTracker::new();
+
+    let mut server = enable_tls(Server::builder(), &config.tls)?;
+
+    info!("Starting cluster-agent on {}", config.address);
```
amorey, line 41: The `info!` macro seems to add the file location and an extra new line by default. Can this be disabled or can we use a different log writer so only the message gets written?

crates/cluster_agent/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "cluster_agent"
+version = "0.1.0"
```
amorey: The version number will come from the release tag so it could be confusing to have one here too. Is it possible to remove it or set it to `0.0.0`?

Tiltfile
```diff
@@ -149,20 +131,24 @@ if build_rust_locally:
...
     deps=[
-      "./crates/rgkl/src",
-      "./crates/rgkl/Cargo.toml",
-      "./crates/rgkl/Cargo.lock",
+      "./crates",
       "./proto",
     ],
+    ignore=[
+      './crates/target',
+      './creates/*/target'
```
amorey: Typo

amorey (changes requested):
Can we remove `modules/cluster-agent` in this PR?

crates/cluster_agent/src/main.rs
gikaragia: So this happens for the 'pretty' formatter, which is meant to be used for development/debugging to be human-readable. In production, we use json by default.

I kind of like the readability when developing. So what we could do instead is to expose the compact option which looks like what you suggest:
```
2025-09-06T10:58:19.901865Z  INFO cluster_agent: Starting cluster-agent on [::]:50051
```

So the possible values of logging.format will be:
- pretty - default in dev environment
- json- default in production
- compact

WDYT?

gikaragia: Also currently we include the filename in json too. Do you want me to remove it?

gikaragia: I implemented like this in 1d6baaee1f64ada226b4cd1478b6b28dfad5729c let me know what you think.

To test it, set the `-p` argument to `logging.format:compact` in Tiltfile

amorey: Ahh ok good to know that there's an option for "compact" too. Rather than introducing a new format option now how about we just stick with the current default behavior (pretty/json) and we revisit this soon with an eye towards syncing logging formats across the cluster components soonish? Maybe if we implement logging with otel libraries then the formats will get standardized automatically.

gikaragia: Sure! Do you want the pretty format to map to compact in order to remove the new line?

amorey: Let’s leave the default pretty behavior (with line number and extra new line) and revisit later

gikaragia: Done in 0ea13a3cbf8fa1f4c424014c415c4a31666ebae5

amorey:
Nice. Can you squash the last commits (5 I think) then I can merge 🚀

amorey (approved):
🦀🥳

PR #639, opened 2025-09-10
✨Add barriers instead of sleeping in tests, use kqueue in macos

### LINKED ISSUE #635: Reduce test run time in cluster agent
# Overview
There is some room for improvement in how fast the tests run. The most important issue is that some of the tests are run serially. 

The reason for that is that when the tests run, they create files in the filesystem, and the files created by one test can affect the other. 

We should investigate if it is possible to make each test isolated and run them in parallel.

gikaragia:
Fixes #635

## Summary

Improves the stability and reduces the run time of Rust tests

## Changes
- Specify a different namespace on each test, to allow the tests to run independently.
- Allow callers of LogMetadataWatcher::watch to specify the watcher type, to allow using stable watchers during testing
- Change all tests apart from `test_create_events_are_generated` to run in parallel. This test isn't possible to run in parallel, as `PollWatcher` throws errors when a tracked file is deleted from another test running in parallel.
- Use barriers or waiting on a precondition instead of sleeping to synchronise threads when testing.
- Use async tasks everywhere instead of creating new threads

## Testing instructions
- Observe that the tests pass.
- Run the tests and observe that the running time is reduced.

## Notes on macOS stability
Generally, the notify `Watcher`s for macOS do not have a consistent behavior, which makes it difficult to test our file watchers on a macOS development environment. Three `Watcher`s exist in macOS:
- `FsEventWatcher`: Optimized to watch whole directories, but it can miss updates on individual files. Good for the LogMetadataService, where the most important part is watching for files getting created/deleted.
- `KqueueWatcher`: Low-level watcher which is good for watching individual files but can miss events on the directories. Good use case for the LogRecordService
- `PollWatcher`: More stable, but it has a performance penalty. Used in testing only.

Currently, in macOS the `RecommendedWatcher` uses `KqueueWatcher`. This means that in order to test consistently file creation scenarios, `PollWatcher` should be used. File deletion scenarios can't be tested consistently because in such a case `PollWatcher` throws errors.

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/log_metadata.rs
+++ b/crates/cluster_agent/src/log_metadata.rs
@@ -1,3 +1,4 @@
+use notify::RecommendedWatcher;
 use prost_types::Timestamp;
 use regex::{Captures, Regex};
 use std::env;
@@ -197,7 +198,7 @@ impl LogMetadataService for LogMetadataImpl {
         );
 
         self.task_tracker.spawn(async move {
-            log_metadata_watcher.watch().await;
+            log_metadata_watcher.watch::<RecommendedWatcher>(None).await;
         });
 
         Ok(Response::new(ReceiverStream::new(log_metadata_rx)))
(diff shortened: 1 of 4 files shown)


amorey (approved):
Wow, nice. You leveled up your Rust.

PR #641, opened 2025-09-12
✨ Improve gRPC stream shutdown handling

amorey:
## Summary

This PR adds support for returning gRPC stream shutdown error codes when `SIGTERM` is received. Previously, streams would shutdown cleanly on `SIGTERM` which meant it wasn't possible to distinguish between a server shutdown and a stream finish. Now clients will receive proper error codes and can react accordingly.

Coded by Codex with human supervision.

## Changes

* Adds a utility wrapper that sends proper error codes to client when shutdown signal is recieved
* Modifies streams to use new wrapper
* Adds unit tests for new wrapper
* Updates CODEOWNERS

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/stream_util.rs
+++ b/crates/cluster_agent/src/stream_util.rs
@@ -1,0 +1,152 @@
+use tokio::select;
+use tokio::sync::{broadcast::Sender as BcSender, mpsc};
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::Status;
+
+/// Wraps a `mpsc::Receiver<Result<T, Status>>` with a termination signal.
+///
+/// When the broadcast channel receives a shutdown signal, this wrapper emits a
+/// single `Err(Status::unavailable("server shutting down"))` and then
+/// terminates the stream. If the inner receiver yields an `Err(Status)`, that
+/// error is forwarded and the stream terminates without emitting an additional
+/// shutdown error. When the inner receiver ends (returns `None`), the stream
+/// completes normally.
+pub fn wrap_with_shutdown<T: Send + 'static>(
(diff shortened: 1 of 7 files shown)


crates/cluster_agent/src/stream_util.rs
```diff
@@ -0,0 +1,152 @@
...
+    let mut term_rx = term_tx.subscribe();
+
+    tokio::spawn(async move {
+        loop {
+            select! {
+                biased;
+                // Prefer shutdown over completing the stream to avoid EOF on shutdown.
+                _ = term_rx.recv() => {
+                    let _ = out_tx
+                        .send(Err(Status::new(tonic::Code::Unavailable, "server shutting down")))
```
gikaragia, line 28: Wouldn't it be simpler if we did this inside the loops of log_metadata_watcher and stream_forward?

amorey: My instinct was to do this upstream from the services because it's a standard gRPC server response and it's easier to test this way. However, maybe each service should decide when/how to handle `SIGTERM`? Do you think the services themselves should handle the response?

gikaragia: Well, the services produce every other response, so I am not sure why this specific one needs special treatment. If there was some complex cleanup that is the same between services, we could investigate sharing the functionality somehow.

But since this seems like just sending a tonic status, I think it will be simpler to do it inside the loops and avoid the extra redirection.

amorey: https://github.com/kubetail-org/kubetail/pull/642

PR #660, opened 2025-09-20
🎣 Patch writer panic in rgkl by ignoring malformed lines

amorey:
## Summary

This PR fixes a panic in the `rgkl` writer that I noticed in development from a malformed log line. Previously we started parsing log strings at the 9th byte ("stdout F"), now we check the string length first and ignore the line if it is less than 9 chars long. It's a strange error because I haven't seen a malformed log line before so it might just a quirk in the dev environment. We should revisit this issue later and decide what to do about malformed log lines. I also noticed an unnecessary use of `.clone()` and `move` in the writer file so I removed those too.

## Changes

* Check string length before slicing in `writer.rs`
* Removed unnecessary use of `sender.clone()` in `writer.rs`
* Removed unnecessary use of `move` in `writer.rs`

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/crates/rgkl/src/util/writer.rs
+++ b/crates/rgkl/src/util/writer.rs
@@ -115,9 +115,8 @@ pub fn process_output(
                                     message: log_msg.trim_end().to_string(),
                                 };
 
-                                let sender = sender.clone();
                                 let result =
-                                    task::block_in_place(move || sender.blocking_send(Ok(record)));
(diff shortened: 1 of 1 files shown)


amorey:
@gikaragia FYI merged small pr to fix panic in rgkl `writer.rs`

PR #685, opened 2025-09-30
🐋 Add timestamps to default output in `kubetail logs`

### LINKED ISSUE #681: Add `TIMESTAMP` to `kubetail logs` output by default
Currently, the `kubetail logs` command just outputs raw log messages without any metadata. We would like to modify the default behavior so that each log event's timestamp is printed in the output (see [parent issue](https://github.com/kubetail-org/kubetail/issues/680) for more context). The timestamp is already available with the `--with-ts` flag but we would like to make this the default and modify the other CLI flags accordingly.

To make this change we need to do the following:

- [x] Make `--with-ts` behavior the default
- [ ] Remove the `--with-ts` flag
- [ ] Add a `--hide-ts` flag
- [ ] Add a `--raw` flag that removes all metadata and only outputs raw messages

The code for the `logs` sub-command is here:
https://github.com/kubetail-org/kubetail/blob/main/modules/cli/cmd/logs.go

jerome-wilson:
Fixes #681 

## Summary

Currently, the default output of `kubetail logs` is just raw log messages. This PR changes the default to include timestamps and also adds a `--raw` flag that removes all metadata and only outputs raw log messages.

## Changes

* Added timestamps to default output
* Removed `--with-ts` flag
* Added `--hide-ts` flag
* Added `--raw` flag

## Submitter checklist

- [x] Rebase branch to HEAD


modules/cli/cmd/logs.go
amorey, line 459: If you use `hideTs` here you can get rid of the `withTs` defined earlier
```suggestion
	if !hideTs {
```

modules/cli/cmd/logs.go
```diff
@@ -215,7 +215,12 @@ var logsCmd = &cobra.Command{
 		archList, _ := flags.GetStringSlice("arch")
 		nodeList, _ := flags.GetStringSlice("node")
 
-		withTs, _ := flags.GetBool("with-ts")
+		hideTs, _ := flags.GetBool("hide-ts")
+		withTs := true
```
amorey: Same comment here regarding `hideTs`. You can modify the code to use `!hideTs` instead of defining a new `withTs`.

modules/cli/cmd/logs.go
```diff
@@ -226,6 +231,18 @@ var logsCmd = &cobra.Command{
 		withContainer, _ := flags.GetBool("with-container")
 		withCursors, _ := flags.GetBool("with-cursors")
 
+		raw, _ := flags.GetBool("raw")
```
amorey, line 232: `--raw` should also set `hideHeader = true`

jerome-wilson: done. can you reverify

amorey (approved):

PR #699, opened 2025-10-04
📜 Update help messages for kubectl plugin usage

### LINKED ISSUE #687: Modify `kubetail` help commands when run as krew plugin
Kubetail was recently added to the krew plugin index:
https://github.com/kubernetes-sigs/krew-index/pull/4892

As per the request of the krew maintainer, we need to modify the `kubetail` help docs to specify `kubectl kubetail` as the CLI name when invoked as a plugin.

wu-seong:
Fixes #687 

## Summary
Updates help messages to display `kubectl kubetail` when running as a kubectl plugin, while maintaining `kubetail` for standalone execution.

## Changes
- Added `CommandDisplayNameAnnotation` to root command for dynamic execution context detection
- Made logs command examples dynamic using `getExamplePrefix()` helper function  
- Added `getCliDisplayName()` to detect kubectl plugin invocation via `os.Args[0]`
- Modified: `modules/cli/cmd/root.go`, `modules/cli/cmd/logs.go`

<details>
<summary>Implementation Notes</summary>

Initially attempted to use `Use: "kubectl kubetail"` directly in the root command, but discovered that Cobra only parses the first word before whitespace in the `Use` field. After investigating [spf13/cobra#2017](https://github.com/spf13/cobra/issues/2017), implemented the solution using `CommandDisplayNameAnnotation`.

</details>

## Submitter checklist
- [x] Link the issue number, if any, to Fixes #687
- [x] Rebase branch to HEAD
- [x] Squash changes into one signed, single commit

### DIFF
--- a/modules/cli/cmd/logs.go
+++ b/modules/cli/cmd/logs.go
@@ -44,7 +44,9 @@ const (
 	logsStreamModeAll
 )
 
-const logsHelp = `
+func getLogsHelp() string {
+	prefix := getExamplePrefix("logs")
+	return fmt.Sprintf(`
 Fetch logs for a specific container or a set of workload containers.
 
 Examples:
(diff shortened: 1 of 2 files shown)


modules/cli/cmd/logs.go
```diff
@@ -44,113 +44,115 @@ const (
 	logsStreamModeAll
 )
 
-const logsHelp = `
+func getLogsHelp() string {
+	prefix := getExamplePrefix("logs")
+	return fmt.Sprintf(`
```
amorey: Instead of using `fmt.Sprintf()`, which requires positional arguments, you can use Go's built-in templating library (`text/template`) to implement it. For example, your template string could look like this:

```go
const logsHelpTmpl = `
Fetch logs for a specific container or a set of workload containers.

Examples:

	- Sources

		# Tail 'web-abc123' pod in 'default' namespace
		{{.CommandDisplayName}} logs web-abc123

		# Tail 'web' deployment in the 'default' namespace
	    {{.CommandDisplayName}} logs deployments/web

...
`
```

modules/cli/cmd/root.go
```diff
@@ -30,19 +32,43 @@ const (
 
 var version = "dev" // default version for local builds
 
+// getCliDisplayName determines the CLI display name based on how it's invoked
+func getCliDisplayName() string {
```
amorey: How about "CommandDisplayName" instead of "CliDisplayName" to match cobra?

modules/cli/cmd/root.go
```diff
@@ -30,19 +32,43 @@ const (
...
+	// Check if running as a kubectl plugin (via krew or direct kubectl invocation)
+	if executable == "kubectl" || strings.HasPrefix(executable, "kubectl-") {
+		return "kubectl kubetail"
+	}
+
+	// Default to standalone binary name
+	return "kubetail"
+}
+
+func getExamplePrefix(subcommand string) string {
```
amorey: What's the advantage of running the subcommand through this function? Can't the sub-commands use `getCommandDisplayName()` and supply their own sub-command strings?

amorey (changes requested):

amorey:
Thanks for your PR and your helpful summary! It looks like the linter check is failing. You can run `gofmt -d -l .` to see what the error is or just run `gofmt -w .` to fix in-place. In case it helps, I have vscode configured to run `gofmt` on save which is very convenient. I also added some other comments in-line.

modules/cli/cmd/root.go
wu-seong: That's a good idea. I'll change it that way.

modules/cli/cmd/logs.go
wu-seong: Thanks for the suggestion! I'll update it to use `text/template` instead of `fmt.Sprintf()`.

modules/cli/cmd/root.go
wu-seong: You're right. Using `getCliDisplayName()` directly is simpler and sufficient. I'll remove the `getExamplePrefix()` helper function.

amorey (approved):
Looks great!

PR #739, opened 2025-10-23
✨Add coverage for newly created file updates in log metadata watcher

### LINKED ISSUE #632: Implement more tests for cluster agent
# Overview
The following areas should have more thorough tests:
- config.rs: No tests at all
- authorizer.rs: No tests at all
- log_metadata_watcher.rs: Add tests that cover updates of a newly created file

itzreze:
Fixes #632

## Summary

This PR adds comprehensive test coverage for the LogMetadataWatcher to verify it correctly handles updates to newly created files. Previously, tests focused on files that existed before the watcher started or on simple create/delete scenarios. These new tests ensure the watcher properly tracks the lifecycle of files that are created and modified during runtime.

## Changes

`test_newly_created_file_receives_modify_events`
- Verifies that a file created after the watcher starts receives both ADDED and MODIFIED events 
- Handles debounced events where create and modify may be combined 
- Tests file size tracking across events

`test_multiple_updates_to_newly_created_file`
- Tests multiple sequential modifications to a newly created file
- Ensures the watcher tracks all modifications with correct file sizes
- Verifies at least one ADDED event and multiple MODIFIED events

`test_create_and_immediate_modify_before_watcher_starts`

- Tests files that exist before the watcher starts
- Ensures only MODIFIED events are generated (not ADDED)
- Verifies proper tracking of pre-existing files

`test_created_file_with_wrong_namespace_not_watched`
- Ensures namespace filtering works for newly created files
- Verifies that files in unwatched namespaces don't generate events
- Tests that modifications to filtered files are ignored

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/crates/cluster_agent/src/log_metadata/log_metadata_watcher.rs
+++ b/crates/cluster_agent/src/log_metadata/log_metadata_watcher.rs
@@ -413,21 +413,20 @@ impl LogMetadataWatchEventType {
 
 #[cfg(test)]
 mod test {
+    use std::fs::{File, remove_file};
+    use std::io::Write;
+
     #[cfg(not(target_os = "macos"))]
-    use std::{
-        fs::{File, remove_file, rename},
-        io::Write,
-    };
+    use std::fs::rename;
+
+    use notify::{PollWatcher, RecommendedWatcher};
+    use serial_test::{parallel, serial};
+    use tokio::sync::{broadcast, mpsc::error::TryRecvError};
+    use tokio::task;
 
     use crate::log_metadata::test::create_test_file;
 
     use super::*;
-    use notify::{PollWatcher, RecommendedWatcher};
-    use serial_test::{parallel, serial};
-    use tokio::{
-        sync::{broadcast, mpsc::error::TryRecvError},
-        task,
-    };
 
     #[tokio::test]
     #[serial]


crates/cluster_agent/src/log_metadata/log_metadata_watcher.rs
amorey, line 419: I'm getting rust-analyzer errors in vscode on macOS because variables that are defined here are out of scope later in the code. Can we put them in scope here or handle the macos exclusion in a different way such that it doesn't cause rust-analyzer/compiler errors?

amorey (changes requested):
Is there a way to re-write the tests using awaits/channels/barriers instead of sleeps? That would speed them up and also make the code easier to follow. For comparison here are some tests that use barriers instead of sleeps:
https://github.com/kubetail-org/kubetail/blob/main/crates/rgkl/src/stream_forward.rs

crates/cluster_agent/src/log_metadata/log_metadata_watcher.rs
```diff
@@ -706,4 +706,359 @@ mod test {
...
+
+        file.write_all(&vec![1; 3])
+            .expect("Failed to write additional data");
+        file.flush().expect("Failed to flush");
+        drop(file);
+
+        sleep(Duration::from_millis(500)).await;
+
+        let mut events = Vec::new();
+        let timeout_duration = Duration::from_secs(5);
```
amorey: Can we run the entire test suite under a timeout rather than handling the timeouts explicitly in the test code?

amorey (changes requested):

itzreze:
@amorey, apologies for the late reply, back then I changed my system configs, my notification bar was messed up probably and I had other commitments to deal with. The PR slipped out of my mind. 
Will try work on this from 19th Nov.
apologies again.

amorey:
No worries, thanks for the update. I'm working on an issue in a separate PR that will require hooking into inner watch events to make clean tests (https://github.com/kubetail-org/kubetail/pull/805) so that might be useful here too. Will let you know.

PR #752, opened 2025-11-02
✨ Add more k8s workload examples to dev environment

### LINKED ISSUE #750: Add more workload examples to the dev environment
We should add an daemonset example and a pod with multiple containers to the dev environment

Babbili:
Fixes #750 

## Summary

This PR adds more k8s workloads examples, _DaemonSet_ and a Multi-containers _Pod_, to the dev environment

## Changes

added 2 k8s resources,
  `DaemonSet` _os-release-daemonset_ 
  `Pod` _multi-containers_

`multi-containers` pod logs 

Screenshot From 2025-11-02 14-57-24

`os-release-daemonset` logs

Screenshot From 2025-11-02 16-30-40

## Submitter checklist

- [ ] Rebase branch to HEAD

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -271,5 +271,3 @@ k8s_resource(
   'echoserver',
   port_forwards='4502:8080',
 )
-
-
--- a/hack/tilt/daemonset-example.yaml
+++ b/hack/tilt/daemonset-example.yaml
@@ -12,15 +12,16 @@ spec:
         app.kubernetes.io/name: daemonset-example
     spec:
       containers:
-      - name: daemonset-example
-        image: alpine:3.22.0
-        env:
-          - name: NODE_NAME
-            valueFrom:
-              fieldRef:
-                fieldPath: spec.nodeName
-        # will print the name of the host machine the pod is running on
-        command: ["sh", "-c"]
-        args: 
-          - echo "This pod is running on $(NODE_NAME)"; sleep infinity
-
+        - name: daemonset-example
+          image: alpine:3.22.0
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          command: ["sh", "-c"]
+          args:
+            - |
+              echo "This pod is running on $(NODE_NAME)"
+              trap "echo 'Received SIGTERM, exiting'; exit 0" TERM
+              sleep infinity & wait $!
--- a/hack/tilt/multi-container-pod.yaml
+++ b/hack/tilt/multi-container-pod.yaml
@@ -3,14 +3,20 @@ kind: Pod
 metadata:
   name: multi-container-pod
 spec:
-  restartPolicy: Never
   containers:
-  - name: container-a
-    image: alpine:3.22.0
-    imagePullPolicy: IfNotPresent
-    command: ["/bin/sh", "-c", "while true; do echo 'Hello from container-a'; sleep 3600; done"]
-  - name: container-b
-    image: alpine:3.22.0
-    imagePullPolicy: IfNotPresent
-    command: ["/bin/sh", "-c", "while true; do echo 'Hello from container-b'; sleep 3600; done"]
-
+    - name: container-a
+      image: alpine:3.22.0
+      command: ["sh", "-c"]
+      args:
+        - |
+          echo "Hello from container-a"
+          trap "echo 'Received SIGTERM, exiting'; exit 0" TERM
+          sleep infinity & wait $!
+    - name: container-b
+      image: alpine:3.22.0
+      command: ["sh", "-c"]
+      args:
+        - |
+          echo "Hello from container-b"
+          trap "echo 'Received SIGTERM, exiting'; exit 0" TERM
+          sleep infinity & wait $!


hack/tilt/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
...
+    matchLabels:
+      app.kubernetes.io/name: os-release
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: os-release
+    spec:
+      containers:
+      - name: log-os-release
+        image: busybox
```
amorey: We use alpine:3.22.0 elsewhere so it would be better to stick with that so k8s can re-use the image

hack/tilt/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: os-release-daemonset
```
amorey: We should attach the `app.kubernetes.com/name` label here too.

hack/tilt/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: os-release
```
amorey: How about `app.kubernetes.io/name: daemonset-example`?

hack/tilt/daemonset.yaml
```diff
@@ -0,0 +1,33 @@
...
+        args:
+          - |
+            cat /etc/os-release
+            while true; do sleep 3600; done
+        volumeMounts:
+        - name: os-release
+          mountPath: /etc/os-release
+          readOnly: true
+      volumes:
+      - name: os-release
```
amorey: Instead of using info from the host, how about printing the name of the node that the container is running on?

hack/tilt/multi-containers-pod.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: multi-containers
```
amorey: How about `name: multi-container-pod`?

hack/tilt/multi-containers-pod.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  restartPolicy: Never
+  volumes:
+  - name: shared-data
+    emptyDir: {}
+  containers:
+  - name: nginx-container
+    image: nginx
```
amorey: Instead of using `nginx` and `debian`, how about using two lighter weight `alpine: 3.22.0` containers?

Tiltfile
```diff
@@ -189,6 +189,8 @@ k8s_yaml('hack/tilt/loggen-ansi.yaml')
 k8s_yaml('hack/tilt/echoserver.yaml')
 k8s_yaml('hack/tilt/cronjob.yaml')
 k8s_yaml('hack/tilt/chaoskube.yaml')
+k8s_yaml('hack/tilt/multi-containers-pod.yaml')
```
amorey: `hack/tilt/multi-container-pod.yaml`

Tiltfile
```diff
@@ -189,6 +189,8 @@ k8s_yaml('hack/tilt/loggen-ansi.yaml')
...
+k8s_yaml('hack/tilt/daemonset.yaml')
```
amorey: `hack/tilt/daemonset-example.yaml`

hack/tilt/daemonset.yaml
amorey: How about `name: daemonset-example`?

hack/tilt/daemonset.yaml
Babbili: sure

hack/tilt/daemonset.yaml
Babbili: that's even better

hack/tilt/multi-containers-pod.yaml
Babbili: yep

hack/tilt/multi-containers-pod.yaml
Babbili: done

Tiltfile
Babbili: done

Tiltfile
Babbili: done

hack/tilt/daemonset-example.yaml
```diff
@@ -0,0 +1,30 @@
...
+        app.kubernetes.io/name: daemonset-example
+    spec:
+      containers:
+      - name: daemonset-example
+        image: alpine:3.22.0
+        # will print the name of the host machine the pod is running on
+        command: ["sh", "-c"]
+        args: 
+          - while true; do echo "This pod is running on $(cat /etc/hostname)"; sleep 3600; done
+        volumeMounts:
```
amorey: Instead of getting the hostname from a volume mount you can add the node name to the env like this:
```
env:
  - name: NODE_NAME
    valueFrom:
      fieldRef:
        fieldPath: spec.nodeName
```

hack/tilt/daemonset-example.yaml
```diff
@@ -0,0 +1,30 @@
...
+      labels:
+        app.kubernetes.io/name: daemonset-example
+    spec:
+      containers:
+      - name: daemonset-example
+        image: alpine:3.22.0
+        # will print the name of the host machine the pod is running on
+        command: ["sh", "-c"]
+        args: 
+          - while true; do echo "This pod is running on $(cat /etc/hostname)"; sleep 3600; done
```
amorey: ChatGPT says that `sh -c` in a linux container runs with PID 1 which doesn't capture `SIGTERM` by default and which might cause issues when shutting down in the while loop. Instead I think this is safer (please verify):

```
args:
  - echo "hi"
  - sleep infinity
```

hack/tilt/daemonset-example.yaml
Babbili: done

PR #757, opened 2025-11-04
✨ add support for --all-containers flag to log subcommand

### LINKED ISSUE #753: Add support for `--all-containers` to `logs` subcommand
Currently, the `kubetail logs` command only returns logs from the main container by default and users must use a container-name wild card to return logs from all containers:

```console
kubetail logs 'deployments/app/*'
```

We should add support for the `--all-containers` flag (which is also used by `kubectl logs`) to enable requests like this:

```console
kubetail logs deployments/frontend deployments/backend --all-containers
```

victorchrollo14:
Fixes #753 

## Summary
Add support for --all-containers to logs subcommand. so when we pass the `--all-containers` flag with the logs command we get logs from all the containers in a Pod.

## Changes
- added a allContainers variable to Sourcewatcher.
- conditional set containerName to "*" if --all-containers is true.

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/logs/source-watcher.go
+++ b/modules/shared/logs/source-watcher.go
@@ -79,14 +79,15 @@ type sourceWatcher struct {
 	cm       k8shelpers.ConnectionManager
 	eventbus evbus.Bus
 
-	kubeContext string
-	bearerToken string
-	regions     []string
-	zones       []string
-	oses        []string
-	arches      []string
-	nodes       []string
-	containers  []string
+	kubeContext   string
+	bearerToken   string
+	regions       []string
+	zones         []string
+	oses          []string
+	arches        []string
+	nodes         []string
+	containers    []string
+	allContainers bool
 
 	allowedNamespaces []string
 	parsedPaths       []parsedPath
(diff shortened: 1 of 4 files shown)


modules/shared/logs/source-watcher.go
```diff
@@ -135,8 +136,14 @@ func NewSourceWatcher(cm k8shelpers.ConnectionManager, sourcePaths []string, opt
 			continue
 		}
 
+		// Show logs from all containers in a pod
+		if sw.allContainers {
+			pp.ContainerName = "*"
```
amorey: Let's make `--all-containers` the default but let users override it with an explicit name. For example users should be able to do this:

```console
kubetail logs deployment/frontend/nginx deployment/backend --all-containers
```

amorey (changes requested):
Looks good! I just made one request to let users override the behavior in the source string.

modules/shared/logs/source-watcher.go
victorchrollo14: Image

victorchrollo14: updated this.

amorey: 🔥

amorey (approved):

PR #776, opened 2025-11-08
feat: Add :latest tag support to Docker images

### LINKED ISSUE #774: Add support for `:latest` tag in docker images
We should also push `:latest` tags when we publish new containers in the `release-docker` workflow (https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/release-docker.yml)

cnaples79:
## Summary
Adds `:latest` tag support to Docker images published in the release workflow, making it easier for users to pull the most recent stable version without specifying a version number.

## Changes Made
Modified `.github/workflows/release-docker.yml` to add `:latest` tags for both Docker Hub and GitHub Container Registry:

- Added two additional `docker buildx imagetools create` commands in the "Create and push manifests" step
- Creates multi-arch `:latest` manifests from the same arch-specific images (amd64, arm64) used for versioned tags
- Applies to both `kubetail/*` and `ghcr.io/*` registries

## How It Works
After creating the versioned multi-arch manifests (e.g., `kubetail/kubetail-dashboard:1.2.3`), the workflow now also creates `:latest` tags pointing to the same images:

```yaml
docker buildx imagetools create -t kubetail/$IMAGE_NAME:latest \
  kubetail/$IMAGE_NAME:$IMAGE_TAG-amd64 \
  kubetail/$IMAGE_NAME:$IMAGE_TAG-arm64
docker buildx imagetools create -t ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:latest \
  ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:$IMAGE_TAG-amd64 \
  ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:$IMAGE_TAG-arm64
```

## Benefits
- Users can now use `docker pull kubetail/kubetail-dashboard:latest` to get the most recent version
- The `:latest` tag is automatically updated with each new release
- Consistent behavior with Docker best practices
- Works for all components (dashboard, cluster-api, cluster-agent)

## Testing
The changes follow the existing pattern for manifest creation and use the same arch-specific images that are already being built and tested.

Fixes #774

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -34,6 +34,7 @@ jobs:
       image_name: ${{ steps.config.outputs.image_name }}
       image_tag: ${{ steps.config.outputs.image_tag }}
       dry_run: ${{ steps.config.outputs.dry_run }}
+      is_latest: ${{ steps.tag_meta.outputs.is_latest }}
     steps:
       - id: config
         run: |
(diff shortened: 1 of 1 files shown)


cnaples79:
Force pushed sign-off commit for DCO check.

amorey (changes requested):
Thanks for the quick pr! This is good for tagging our most recent releases as `latest` but we should also add an exclusion for beta releases (e.g. `0.8.1-rc-1`). I think it's probably ok to only apply `latest` when the tag has all-numeric semver format.

Also - I don't know how difficult it would be but it would be cool if we could retrieve the previous `latest` and only apply the tag to the new one if its version is greater. This might take a while so no worries if you want to leave it out of this pr.

amorey:
@cnaples79 Just checking in - do you think you'll have time to make those changes to the `:latest` handling? If not, let me know. There's another contributor interested in working on this but he can base his changes on your work so you'll still get credit.

PR #801, opened 2025-11-14
🎣 Fix stale in-cluster ServiceAccount token issue

amorey:
## Summary

This PR fixes an issue causing unauthorized connection issues between the in-cluster dashboard and the cluster-api. Previously, we assumed that the standard token which k8s mounts locally had no expiration but this is not the case, at least not for some installations. Now the cluster-api proxy in the dashboard uses a transport that automatically refreshes the SA token if it is close to expiring.

## Changes

* Implement `InClusterSATRoundTripper` and use in dashboard cluster-api proxy

## Submitter checklist

- [x] Rebase branch to HEAD


amorey:
@codex review

PR #806, opened 2025-11-16
🐋 Add Docker image support for CLI component

### LINKED ISSUE #789: Publish a docker image for the `kubetail` CLI tool
To support users who want to use Kubetail locally running inside a docker container as well as users who want to use the `kubetail` CLI tool inside their clusters or in custom docker images we should create a `kubetail-cli` docker image and integrate it into our CLI release workflow.

The `kubetail-cli` docker image should have the following features:
* The `kubetail` binary should be static
* The base image should be based on busybox and we should publish these tags/variants:
    ```
    kubetail-cli:${VERSION}
    kubetail-cli:0.10.0-alpine
    kubetail-cli:latest
    kubetail-cli:latest-alpine
    ```

In addition:
* We should publish the images to docker hub and to ghcr.io
* We should sign the images using cosign

Currently the CLI release workflow is located in [`.github/workflows/release-cli.yml`](https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/release-cli.yml) but I think it might be better to put the CLI docker-specfic release workflow in [`.github/workflows/release-docker.yml`](https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/release-docker.yml).

smazmi:
Fixes #789

## Summary

This PR adds Docker image publishing support for the kubetail CLI tool with two base image variants (busybox and alpine). Images will be published to both Docker Hub and GHCR with cosign signing.

## Changes

- **New: Dockerfile.cli** - Multi-stage Dockerfile with frontend builder, Go builder, and two final stages (busybox and alpine variants)
  - Static binary compilation with `CGO_ENABLED=0` for portability
  - Embeds dashboard frontend for `kubetail serve` command

- **New: .github/workflows/release-docker-cli.yml** - Dedicated release workflow for CLI component
  - Triggers on `cli/v*` tags
  - Builds both busybox (default) and alpine variants in parallel
  - Creates four manifest tags: `VERSION`, `latest`, `VERSION-alpine`, `latest-alpine`
  - Signs all four variants to both Docker Hub and GHCR with cosign
  - Separate workflow keeps logic simple and avoids conditionals in main release-docker.yml

## Usage

```bash
# Busybox variant (default)
docker run -v ~/.kube:/root/.kube:ro kubetail/kubetail:latest logs <pod-name>

# Alpine variant
docker run -v ~/.kube:/root/.kube:ro kubetail/kubetail:latest-alpine logs <pod-name>
```

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/.github/workflows/release-docker-cli.yml
+++ b/.github/workflows/release-docker-cli.yml
@@ -1,0 +1,293 @@
+name: release-docker-cli
+
+permissions:
+  contents: read
+  packages: write
+  deployments: write
+  id-token: write
+
+on:
+  push:
+    tags:
+      - "cli/v*"
+  workflow_dispatch:
+    inputs:
+      git_tag:
+        description: "Git tag"
+        required: true
+        type: string
+      dry_run:
+        description: "Dry run"
+        required: false
+        type: boolean
+        default: true
+
+jobs:
+  config:
+    name: Configure workflow
+    runs-on: ubuntu-24.04
+    outputs:
+      git_tag: ${{ steps.config.outputs.git_tag }}
+      component: ${{ steps.config.outputs.component }}
+      image_name: ${{ steps.config.outputs.image_name }}
+      image_tag: ${{ steps.config.outputs.image_tag }}
+      dry_run: ${{ steps.config.outputs.dry_run }}
+    steps:
+      - id: config
+        run: |
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            GIT_TAG="${{ github.event.inputs.git_tag }}"
+            DRY_RUN="${{ github.event.inputs.dry_run }}"
+          else
+            GIT_TAG="${GITHUB_REF#refs/tags/}"
+            DRY_RUN=false
+          fi
+
+          COMPONENT=${GIT_TAG%%/*}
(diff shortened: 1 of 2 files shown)


amorey (changes requested):
This looks great, thanks. This PR adds cli-specific steps to the `release-docker` workflow. Instead, to keep things simple for the follow-up PR where we add `-alpine` tags to the other components, how about we put the CLI builder into it's own separate workflow (e.g. `release-docker-cli.yml`)? Then in the follow-up PR we can merge the workflows back into one file without `is_cli` exceptions.

smazmi:
> This looks great, thanks. This PR adds cli-specific steps to the `release-docker` workflow. Instead, to keep things simple for the follow-up PR where we add `-alpine` tags to the other components, how about we put the CLI builder into it's own separate workflow (e.g. `release-docker-cli.yml`)? Then in the follow-up PR we can merge the workflows back into one file without `is_cli` exceptions.

Thanks for the feedback @amorey! Yeah, that sounds like the cleanest approach to me too. Once we have the rest of the Alpine builds in place, consolidating the two workflows back into a single file will be much simpler. I'll go ahead and revert `release-docker.yml` to match the latest commit and move the CLI release logic into a separate `release-docker-cli.yml` workflow for now.

PR #812, opened 2025-11-17
🐋 Add alpine variants for all components and consolidate workflows

### LINKED ISSUE #810: Add alpine image variants for dashboard/cluster-api/cluster-agent and consolidate release workflows
Currently, only the CLI component has alpine-based Docker image variants (added in #806). We should extend this to all other components and consolidate the release workflows.

### Current State
- **CLI**: Has both busybox and alpine variants (`Dockerfile.cli`)
  - Released via separate `release-docker-cli.yml` workflow
- **Dashboard, Cluster-API, Cluster-Agent**: Only have busybox-based images
  - Released via `release-docker.yml` workflow

#### Part 1: Add Alpine Variants
Update the following Dockerfiles to include alpine variants (similar to `Dockerfile.cli`):
- `build/package/Dockerfile.dashboard` - add `final-alpine` target
- `build/package/Dockerfile.cluster-api` - add `final-alpine` target  
- `build/package/Dockerfile.cluster-agent` - add `final-alpine` target

#### Part 2: Consolidate Release Workflows
Once alpine variants exist for all components:
- Update `release-docker.yml` to build alpine variants for all components
- Remove the temporary `release-docker-cli.yml` workflow
- Ensure all components follow the same tagging and signing patterns

### References
- CLI implementation: `build/package/Dockerfile.cli`
- Current workflow: `.github/workflows/release-docker.yml`
- Temporary CLI workflow: `.github/workflows/release-docker-cli.yml`
- Related PR: #806

smazmi:
Fixes #810

## Summary

This PR extends alpine image variant support to all Docker components (dashboard, cluster-api, cluster-agent) and consolidates the release workflows. Previously, only the CLI had alpine variants (added in #806) with a separate workflow. Now all components follow a unified approach with a single consolidated workflow.

## Changes

- **Updated `build/package/Dockerfile.cli`** - Renamed `final-busybox` to `final` for consistency
  - Now follows the same naming convention as other components (final = default, final-alpine = alpine variant)
  - Simplifies workflow logic by eliminating special-case handling

- **Updated `build/package/Dockerfile.dashboard`** - Added `final-alpine` stage
  - Mirrors the `final` (scratch) stage but uses alpine:3.22.0 as base
  - Includes SSL certificates for TLS verification
  - Same binary from dashboard-builder stage

- **Updated `build/package/Dockerfile.cluster-api`** - Added `final-alpine` stage  
  - Mirrors the `final` (scratch) stage but uses alpine:3.22.0 as base
  - Simple static binary copy (no dependencies needed)

- **Updated `build/package/Dockerfile.cluster-agent`** - Added `final-alpine` stage
  - Uses alpine:3.22.0 with `libgcc` package for Rust glibc compatibility
  - Includes grpc-health-probe binary
  - Simpler than scratch variant (no manual library copying needed)

- **Updated `.github/workflows/release-docker.yml`** - Consolidated and simplified for all components
  - Added `cli/v*` tag trigger (consolidates CLI releases)
  - Removed conditional target selection - all components now use consistent `final` and `final-alpine` targets
  - Split build step into "default" and "alpine" variants
  - Split manifest creation into separate default and alpine steps
  - Updated digest fetching to handle 4 variants (default, alpine, latest, latest-alpine)
  - Updated cosign signing to sign all variants
  - Handles conditional latest tag logic with `IS_LATEST` env var

- **Updated `.github/workflows/ci.yml`** - Added alpine variant tests
  - Renamed `cli-busybox` to `cli-final` for consistency with other components
  - Added `dashboard-alpine` build test (target: final-alpine)
  - Added `cluster-api-alpine` build test (target: final-alpine)
  - Added `cluster-agent-alpine` build test (target: final-alpine)
  - All tested on both ubuntu-24.04 (amd64) and ubuntu-24.04-arm (arm64)

- **Removed `.github/workflows/release-docker-cli.yml`** - No longer needed
  - CLI now handled by consolidated `release-docker.yml`

## Image Tags

Each component now publishes 4 image tags (or 2 if not latest):
- `VERSION` - default variant (busybox for CLI, scratch for others)
- `latest` - default variant (only if is_latest)
- `VERSION-alpine` - alpine variant
- `latest-alpine` - alpine variant (only if is_latest)

All components now use consistent stage naming: `final` (default) and `final-alpine` (alpine variant).

Example for dashboard v1.2.3 (if latest):
- `kubetail/kubetail-dashboard:1.2.3`
- `kubetail/kubetail-dashboard:latest`
- `kubetail/kubetail-dashboard:1.2.3-alpine`
- `kubetail/kubetail-dashboard:latest-alpine`

## Architecture

The changes maintain consistency across all components:

| Component      | Default Base | Alpine Base |
|---------------|--------------|-------------|
| dashboard     | scratch      | alpine:3.22.0 |
| cluster-api   | scratch      | alpine:3.22.0 |
| cluster-agent | scratch      | alpine:3.22.0 |
| cli           | busybox      | alpine:3.22.0 |

All images support multi-arch (amd64, arm64) and are signed with cosign.

## Submitter checklist

- [x] Rebase branch to HEAD


.github/workflows/release-docker.yml
```diff
@@ -178,10 +192,38 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:$IMAGE_TAG-arm64
             
           if [ "$IS_LATEST" = "true" ]; then
+            # Create latest manifest (default/scratch)
             docker buildx imagetools create -t kubetail/$IMAGE_NAME:latest \
-              kubetail/$IMAGE_NAME:$IMAGE_TAG
+              kubetail/$IMAGE_NAME:$IMAGE_TAG-amd64 \
```
amorey: I think aliasing the `:latest` tag to `:$IMAGE_TAG` is simpler and easier to maintain. Is there a benefit to mapping it to `-amd64` and `-arm64` again?

.github/workflows/release-docker.yml
```diff
@@ -178,10 +192,38 @@ jobs:
...
+              kubetail/$IMAGE_NAME:$IMAGE_TAG-arm64
             docker buildx imagetools create -t ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:latest \
-              ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:$IMAGE_TAG
+              ghcr.io/${{ github.repository_owner }}/$IMAGE_NAME:$IMAGE_TAG-amd64 \
```
amorey: Same as above

build/package/Dockerfile.cluster-agent
```diff
@@ -84,3 +84,21 @@ COPY --from=builder /work/crates/target/release/cluster_agent cluster-agent
...
+FROM alpine:3.22.0 AS final-alpine
+
+WORKDIR /cluster-agent
+
+# install runtime dependencies (glibc compatibility)
+RUN apk add --no-cache libgcc
+
+# copy grpc-health-probe
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.38 /ko-app/grpc-health-probe /bin/grpc_health_probe
+
```
amorey, line 96: The cluster-agent executable is compiled against glibc (for performance reasons) but Alpine uses musl so you also need to copy the glibc runtime dependencies for it to work:

```suggestion
# copy glibc runtime libraries and dynamic loader (aarch64)
COPY --from=builder /lib/aarch64-linux-gnu/libgcc_s.so.1 /lib/aarch64-linux-gnu/
COPY --from=builder /lib/aarch64-linux-gnu/libc.so.6 /lib/aarch64-linux-gnu/
COPY --from=builder /lib/aarch64-linux-gnu/libm.so.6 /lib/aarch64-linux-gnu/
COPY --from=builder /lib/ld-linux-aarch64.so.1 /lib/

# copy glibc runtime libraries and dynamic loader (x86_64)
COPY --from=builder /lib/x86_64-linux-gnu/libgcc_s.so.1 /lib/x86_64-linux-gnu/
COPY --from=builder /lib/x86_64-linux-gnu/libc.so.6 /lib/x86_64-linux-gnu/
COPY --from=builder /lib/x86_64-linux-gnu/libm.so.6 /lib/x86_64-linux-gnu/
COPY --from=builder /lib64/ld-linux-x86-64.so.2 /lib64/

```

amorey (changes requested):
Overall looks good. I added some comments and suggestions. Did you consider using a matrix strategy in `release-docker.yml` to avoid duplicating code?

.github/workflows/release-docker.yml
smazmi: Oh, my bad, my thought process actually was: since we're creating the `$IMAGE_TAG` manifest from the arch-specific tags (`-amd64`, `-arm64`), we should do the same for `latest` to keep it consistent and direct, that way both manifests are created the same way and point directly to the architecture images.

But now in hindsight, previous approach is actually cleaner and simpler, the `$IMAGE_TAG` manifest already exists and points to the right arch images. Same result, less redundancy.
I'll revert it back to aliasing. Thanks for the feedback

build/package/Dockerfile.cluster-agent
smazmi: Understood, I will add them in

amorey:
@smazmi The last commit was a merge commit from `main` into your branch which creates a non-linear history. Can you undo the last commit and do a rebase instead?

amorey:
Merged. Thanks! 🥳

PR #833, opened 2025-11-28
✨Improve CLI tool feedback format

### LINKED ISSUE #821: Improve CLI tool feedback format
Currently the CLI tool uses the normal logging route to send messages back to the user which formats messages using a log line format by default. For example, here's the response when the `kubeconfig` can't be found:

```console
$ kubetail serve --kubeconfig /path/does/not/exist
2025-11-21T10:27:17.749458+03:00 FTL modules/cli/cmd/serve.go:127 > error="kubeconfig file not found at '/path/does/not/exist'.\n\nPlease ensure the file exists or use the '--kubeconfig' flag to specify a custom path.\nIf you are running inside a cluster, use the '--in-cluster' flag"
```

Although the log line format is useful for some CLI features (e.g. `kubetail serve`) in most cases it would be better to return human-formatted strings like this:

```console
$ kubetail serve --kubeconfig /path/does/not/exist
Error: kubeconfig file not found at "/path/does/not/exist"

Ensure the file exists or provide a valid path using the --kubeconfig flag.
If running inside a Kubernetes cluster, use the --in-cluster flag instead.
```

We should change the default logger formatter for the CLI tool so that messages are returned directly to the user without being log line formatted. In addition it would be nice to prefix the first line of the message with the log level (e.g. `Error:`, `Warn:`).

DavidReque:
Fixes #821

## Summary
Add new `cli` logger format that displays user-friendly messages
without timestamps, caller info, or log line formatting.

## Changes
- Add `cli` format option to `ConfigureLogger` in `config.go`
- Format displays `"Error:" prefix for fatal/error` levels
- Format displays `"Warn:" prefix for warning` levels
- No prefix for info messages
- Update `serve.go` to use `cli` format instead of `pretty`

## Submitter checklist
- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/config/config.go
+++ b/modules/shared/config/config.go
@@ -493,9 +493,9 @@ func ConfigureLogger(opts LoggerOptions) {
 					}
 					switch i.(string) {
 					case "fatal", "error":
-						return "\033[31mError:\033[0m"
+						return "\033[31mError:\033[0m "
 					case "warn":
-						return "\033[33mWarn:\033[0m"
+						return "\033[33mWarn:\033[0m "
 					default:
 						return ""
 					}
@@ -507,7 +507,7 @@ func ConfigureLogger(opts LoggerOptions) {
 					if i == nil {
 						return ""
 					}
-					return fmt.Sprintf(" %s", i)
+					return fmt.Sprintf("%s", i)
 				},
 				FormatFieldName: func(i interface{}) string {
 					return ""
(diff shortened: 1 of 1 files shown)


amorey (changes requested):
This is great. One small suggestion - currently `FormatMessage()` adds a space before the string so fields without a level prefix look like this:
```
$ kubetail serve
 Started kubetail dashboard on http://localhost:7500/
```

Instead, how about appending a space in `FormatLevel()` when there's a prefix so we can remove the empty space when there's no prefix?

amorey (approved):

Issue #859, opened 2025-12-14
Add support for CLI config file


amorey:
### Overview

We would like to add support for a CLI config file so that users can configure their default settings when using CLI commands. For example, users should be able to set the default columns they would like to see when tailing logs or the default port number to use when running the web dashboard.

### Goals

* [ ] Allow users to set defaults for `logs` sub-command
* [ ] Allow users to set defaults for `serve` sub-command
* [ ] By default look for config file in `~/.config/kubetail/config`
* [ ] Add a global flag to the CLI tool to support custom config file paths (e.g. `-c, --config`)
* [ ] Support yaml by default with other formats as indicated by file extension

### Non-Goals

* We don't need to support multiple config file versions
* We don't need to support features beyond the ones listed above

### Design Overview

We want to support the following config schema:

```yaml
## Kubetail CLI Configuration
#

## version ##
#
# Sets the version of the config schema
#
# Default value: 1
#
version: 1

## general ##
#
general:

  ## kubeconfig ##
  #
  # Overrides the default kubeconfig path ($HOMEDIR/.kube/config)
  #
  # Default value: __empty__
  #
  kubeconfig:

## commands ##
#
commands:

  ## logs ##
  #
  logs:

    ## head ##
    #
    # Sets the default number of log records to return for `head`
    #
    # Default value: 10
    #
    head: 10

    ## tail ##
    #
    # Sets the default number of log records to return for `tail`
    #
    # Default value: 10
    #
    tail: 10

    ## columns ##
    #
    # Sets the default columns to show when displaying log records
    #
    # Default value: [ts, dot, message]
    #
    columns:
      - ts
      - dot
      - message

  ## serve ##
  #
  serve:

    ## port ##
    #
    # The port to use for the dashboard server
    # 
    # Default: 7500
    #
    port: 7500
```

### Future Work

Here are some ideas for other features we can support in the future:
https://github.com/kubetail-org/cli-config-proposal/blob/main/config.yaml

PR #867, opened 2025-12-19
🐋 Add support for CLI config file

rf-krcn:
Fixes #https://github.com/kubetail-org/kubetail/issues/859

## Summary

This PR adds support for CLI configuration files, allowing users to define persistent defaults for CLI sub-commands. Currently we are only adding support for `logs` and `serve` sub-commands, as well as a default kubeconfig path, via ~/.config/kubetail/config or a custom --config flag.

## Changes

* Adds new CLI config struct
* Adds `NewCLIConfigFromFile()` and `NewCLIConfigFromViper()` methods

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/cli/cmd/logs.go
+++ b/modules/cli/cmd/logs.go
@@ -647,7 +647,10 @@ func init() {
 	// is called directly, e.g.:
 	// serveCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	defaultConfigPath, _ := config.DefaultConfigPath()
-	cliCfg := config.NewCLIConfigFromFile(defaultConfigPath)
+	cliCfg, err := config.NewCLIConfigFromFile(defaultConfigPath)
+	if err != nil {
+		cliCfg = config.DefaultCLIConfig()
+	}
 
 	flagset := logsCmd.Flags()
 	flagset.SortFlags = false
--- a/modules/shared/config/cli.go
+++ b/modules/shared/config/cli.go
@@ -70,10 +70,10 @@ func DefaultConfigPath() (string, error) {
 	return filepath.Join(home, ".kubetail", "config.yaml"), nil
 }
 
-func NewCLIConfigFromFile(configPath string) *CLIConfig {
+func NewCLIConfigFromFile(configPath string) (*CLIConfig, error) {
 	if configPath == "" {
 		if f, err := DefaultConfigPath(); err != nil {
-			return nil
+			return nil, err
 		} else {
 			configPath = f
 		}
@@ -84,20 +84,18 @@ func NewCLIConfigFromFile(configPath string) *CLIConfig {
 	// read contents
 	configBytes, err := os.ReadFile(configPath)
 	if err == nil {
-		//		return nil, err
-
 		// expand env vars
 		configBytes = []byte(os.ExpandEnv(string(configBytes)))
 
 		// check extension
 		if len(filepath.Ext(configPath)) <= 1 {
-			return nil
+			return nil, fmt.Errorf("file %q must have a valid extension (e.g., .yaml, .json)", configPath)
 		}
 
 		// load into viper
 		v.SetConfigType(filepath.Ext(configPath)[1:])
 		if err := v.ReadConfig(bytes.NewBuffer(configBytes)); err != nil {
-			return nil
+			return nil, err
 		}
 	}
 
@@ -105,16 +103,15 @@ func NewCLIConfigFromFile(configPath string) *CLIConfig {
 
 	// unmarshal
 	if err := v.Unmarshal(cfg); err != nil {
-		return nil
+		return nil, err
 	}
 
 	// validate config
 	if err := cfg.validate(); err != nil {
-		return nil
+		return nil, err
 	}
 
-	return cfg
-
+	return cfg, nil
 }
 
 func NewCLIConfigFromViper(v *viper.Viper, configPath string) (*CLIConfig, error) {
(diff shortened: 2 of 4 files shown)


modules/cli/cmd/logs.go
```diff
@@ -626,9 +655,9 @@ func init() {
 
 	flagset.String(KubeContextFlag, "", "Specify the kubeconfig context to use")
 	flagset.Int64P("head", "h", 10, "Return first N records")
-	flagset.Lookup("head").NoOptDefVal = "10"
+	flagset.Lookup("head").NoOptDefVal = "-1"
```
amorey: With "-1" the help docs look like this:
```
  -h, --head int[=-1]         Return first N records (default 10)
  -t, --tail int[=-1]         Return last N records (default 10)
```

How about we use "N" instead? I think that will make `--help` more presentable.

modules/cli/cmd/serve.go
```diff
@@ -207,39 +218,41 @@ var serveCmd = &cobra.Command{
...
 	// Init viper
 	v := viper.New()
-	v.BindPFlag("dashboard.logging.level", cmd.Flags().Lookup("log-level"))
-	v.Set("dashboard.addr", fmt.Sprintf("%s:%d", host, port))
 
+	v.BindPFlag("general.kubeconfig", cmd.Flags().Lookup(KubeconfigFlag))
+	v.BindPFlag("commands.serve.port", cmd.Flags().Lookup("port"))
+	v.BindPFlag("commands.serve.host", cmd.Flags().Lookup("host"))
+	v.BindPFlag("commands.serve.skip-open", cmd.Flags().Lookup("skip-open"))
+	v.BindPFlag("dashboard.logging.level", cmd.Flags().Lookup("log-level"))
```
amorey: There's a typo here but I think we can skip `log-level` altogether for now. Let's launch the config file feature with `port/host/skip-open` and then see which other settings users want to control from the config file.

modules/shared/config/cli.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"os"
+
+	"path/filepath"
+
+	"github.com/go-playground/validator/v10"
+	"github.com/spf13/viper"
+)
+
+const ConfigPath = "$HOME/.config/kubetail/config.yaml"
```
amorey: To do this in a safe cross-platform way we should get the homedir from Go. Also, I keep going back-and-forth about whether it should go in `~/.config/kubetail` or `~/.kubetail` but to be consistent with other Kubernetes tools let's use `~/.kubetail`:

```go
import (
	"fmt"
	"os"
	"path/filepath"
)

func DefaultConfigPath() (string, error) {
	home, err := os.UserHomeDir()
	if err != nil {
		return "", fmt.Errorf("get home dir: %w", err)
	}

	return filepath.Join(home, ".kubetail", "config.yaml"), nil
}
```

modules/shared/config/cli.go
```diff
@@ -0,0 +1,108 @@
...
+type CLI struct {
+	Config string `validate:"omitempty,file"`
+}
+
+// Application configuration
+type CLIConfig struct {
+	// Global settings
+	General struct {
+		KubeconfigPath string `mapstructure:"kubeconfig"`
+		string         `mapstructure:"kubeconfig"`
```
amorey:
```suggestion
```

modules/shared/config/cli.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"path/filepath"
+
+	"github.com/go-playground/validator/v10"
+	"github.com/spf13/viper"
+)
+
+const ConfigPath = "$HOME/.config/kubetail/config.yaml"
+
+type CLI struct {
```
amorey, line 14: Is this struct being used anywhere?

modules/shared/config/cli.go
```diff
@@ -0,0 +1,108 @@
...
+			KubeContext string `mapstructure:"kube-context"`
+			Head        int64  `mapstructure:"head"`
+			Tail        int64  `mapstructure:"tail"`
+		} `mapstructure:"logs"`
+
+		// Default behavior for the 'serve' command
+		Serve struct {
+			Port     int    `mapstructure:"port"`
+			Host     string `mapstructure:"host"`
+			LogLevel string `mapstructure:"log-level"`
```
amorey:
```suggestion
```

modules/shared/config/cli.go
```diff
@@ -0,0 +1,108 @@
...
+func DefaultCLIConfig() *CLIConfig {
+	cfg := &CLIConfig{}
+
+	cfg.Commands.Logs.KubeContext = ""
+	cfg.Commands.Logs.Head = 10
+	cfg.Commands.Logs.Tail = 10
+
+	cfg.Commands.Serve.Port = 7500
+	cfg.Commands.Serve.Host = "localhost"
+	cfg.Commands.Serve.LogLevel = "info"
```
amorey:
```suggestion
```

amorey (changes requested):
Looks great! I've added some suggestions.

amorey:
The `crates-test` failure is from an intermittent rate limiting issue and will go away on its own. I think the `modules-test` check will pass when you revert the `log-level` behavior.

modules/cli/cmd/logs.go
rf-krcn: It won't work if we do that because `NoOptDefVal` only accepts integers as default values in here.

modules/cli/cmd/serve.go
rf-krcn: It is easy to fix, `dashboard.logging.level` should have been `commands.serve..log-level`

modules/shared/config/cli.go
rf-krcn: It is used inside `PreRunE` in both `logs.go` and `serve.go` with the validator.

amorey:
@codex review

PR #868, opened 2025-12-29
✨ Add basic tests for CLI `logs` sub-command

victorchrollo14:
## Summary

This PR moves the logs code into runLogs so we can add unit tests and extracted the cm and stream part so we can pass in mock data while testing.

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/cli/cmd/logs.go
+++ b/modules/cli/cmd/logs.go
@@ -27,10 +27,8 @@ import (
 	"time"
 
 	"github.com/go-playground/validator/v10"
-	zlog "github.com/rs/zerolog/log"
 	"github.com/sosodev/duration"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
 	sharedcfg "github.com/kubetail-org/kubetail/modules/shared/config"
@@ -53,6 +51,51 @@ const (
 	logsStreamModeTail
 	logsStreamModeAll
 )
+
+type logsConfig struct {
+	streamOpts []logs.Option
+
+	kubecontext    string
+	inCluster      bool
+	kubeconfigPath string
+
+	sinceTime time.Time
+	untilTime time.Time
+
+	head    bool
+	headVal int64
+	tail    bool
+	tailVal int64
+	all     bool
+	follow  bool
+
+	grep       string
+	regionList []string
+	zoneList   []string
+	osList     []string
+	archList   []string
+	nodeList   []string
+
+	hideHeader bool
+	hideTs     bool
+	hideDot    bool
+
+	withTs        bool
+	withDot       bool
+	allContainers bool
+
+	withNode      bool
+	withRegion    bool
+	withZone      bool
+	withOS        bool
+	withArch      bool
+	withNamespace bool
+	withPod       bool
+	withContainer bool
+	withCursors   bool
+
+	raw bool
+}
 
 const logsHelpTmpl = `
 Fetch logs for a specific container or a set of workload containers.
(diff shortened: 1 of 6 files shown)


modules/cli/cmd/logs.go
```diff
@@ -147,7 +147,7 @@ Examples:
 		{{.CommandDisplayName}} deployments/web --region=us-east-1,us-east-2
 
 		# Tail 'web' deployment pods in 'us-east-1' running on 'arm64'
-		{{.CommandDisplayName}} deployments/web --region=us-east-1 --arch=arm64
+		 {{.CommandDisplayName}} deployments/web --region=us-east-1 --arch=arm64
```
amorey:
```suggestion
		{{.CommandDisplayName}} deployments/web --region=us-east-1 --arch=arm64
```

modules/cli/cmd/logs.go
```diff
@@ -609,19 +623,8 @@ func orDefault[T comparable](val T, defaultVal T) T {
...
-
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// serveCmd.PersistentFlags().String("foo", "", "A help for foo")
-
-	// Cobra supports local flags which will only run when this command
-	// is called directly, e.g.:
-	// serveCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
-	flagset := logsCmd.Flags()
+func AddLogsCmdFlags(cmd *cobra.Command) {
```
amorey: This doesn't need to be public so we can name it `addLogsCmdFlags()`

modules/cli/cmd/logs.go
```diff
@@ -193,266 +199,274 @@ func getLogsHelp() string {
...
-var logsCmd = &cobra.Command{
-	Use:   "logs [source1] [source2] ...",
-	Short: "Fetch logs for a container or a set of workloads",
-	Long:  strings.ReplaceAll(getLogsHelp(), "\t", "  "),
-	Args:  cobra.MinimumNArgs(1),
-	PreRunE: func(cmd *cobra.Command, args []string) error {
-		flags := cmd.Flags()
-		grep, _ := flags.GetString("grep")
-		force, _ := flags.GetBool("force")
+func RunLogs(cmd *cobra.Command, args []string, createStream StreamFactory, cm k8shelpers.ConnectionManager) {
```
amorey: This should be private (`runLogs`)

modules/cli/cmd/logs_test.go
```diff
@@ -0,0 +1,382 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			cmd := &cobra.Command{}
+			addLogsCmdFlags(cmd)
+
+			test.setup(cmd)
+			flags, streamOpts := loadLogConfig(cmd)
+
+			assert.NotEmpty(t, streamOpts)
+			assert.Len(t, streamOpts, 12)
```
victorchrollo14: the streamOpts doesn't have a proper test. @amorey  should we apply this options to a stream struct and then verify the expected data here or is this sufficient to test?

modules/shared/logs/stream.go
```diff
@@ -29,6 +29,14 @@ import (
...
 const DEFAULT_MAX_CHUNK_SIZE = 16 * 1024 // 16 KB
 
+// Streamer interface defines the methods for consuming a stream
+type Streamer interface {
+	Sources() []LogSource
+	Records() <-chan LogRecord
+	Err() error
+	Close()
+}
+
```
victorchrollo14, line 39: llm suggested to use Streamer instead of `Stream` for interface and `stream` for struct, is this alright or better to follow the approach you suggested earlier.

modules/shared/logs/mock/stream.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2024-2025 Andres Morey
```
amorey:
```suggestion
// Copyright 2024 The Kubetail Authors
```

modules/cli/cmd/logs.go
```diff
@@ -51,6 +50,46 @@ const (
 	logsStreamModeAll
 )
 
+type logFlags struct {
```
amorey: The use of name `Flags` here is confusing because we use pflags elsewhere. How about `logsOptions` or something similar?

modules/cli/cmd/logs.go
```diff
@@ -197,187 +236,324 @@ func getLogsHelp() string {
...
-var logsCmd = &cobra.Command{
-	Use:   "logs [source1] [source2] ...",
-	Short: "Fetch logs for a container or a set of workloads",
-	Long:  strings.ReplaceAll(getLogsHelp(), "\t", "  "),
-	Args:  cobra.MinimumNArgs(1),
-	PreRunE: func(cmd *cobra.Command, args []string) error {
-		flags := cmd.Flags()
-		grep, _ := flags.GetString("grep")
-		force, _ := flags.GetBool("force")
+func loadLogConfig(cmd *cobra.Command) (*logFlags, []logs.Option) {
```
amorey: Instead of returning `(*logFlags, []logs.Option)` I'd recommend returning `(*logsOptions, error)` (with `[]logs.Option` embedded in `logsOptions`). Otherwise the function will handle errors internally which makes it harder to test.

modules/cli/cmd/logs.go
```diff
@@ -197,187 +236,324 @@ func getLogsHelp() string {
...
+	}
+
+	// Default tail num to 0 if follow is true
+	if follow && !tail {
+		tailVal = 0
+	}
+
+	// Parse `since`
+	sinceTime, err := parseTimeArg(since)
+	cli.ExitOnError(err)
```
amorey: See above about returning an error instead of exiting the program here

modules/cli/cmd/logs_test.go
```diff
@@ -0,0 +1,382 @@
...
+	"testing"
+	"time"
+
+	"github.com/kubetail-org/kubetail/modules/shared/logs"
+	stream "github.com/kubetail-org/kubetail/modules/shared/logs/mock"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+func defaultLogFlags() logFlags {
```
amorey: Introducing the "defaults" here isn't accurate because the defaults are actually defined in the flags themselves so this introduces another place the code can drift from the tests. Here are some ideas for how to improve the code and the tests:

* Put the flag defaults into a struct that can be imported by `addLogsCmdFlags()` and the tests
* Separate flag parsing from CLIConfig parsing so you can test one without the other
* Focus the tests on flag parsing that triggers business logic (e.g. `--after`/`--before`) instead of testing defaults

modules/cli/cmd/logs_test.go
amorey: Lets's simplify the other tests (see above) then return to this afterwards.

modules/shared/logs/stream.go
amorey: Hmm GitHub duplicated my previous comment and then deleted both of them when I only wanted to delete one 😫. Let's use `type Stream` for now and revisit in the future (`Streamer` doesn't make sense with our current instance names like `stream logs.Streamer`)

modules/cli/cmd/logs.go
```diff
@@ -566,46 +663,46 @@ func getTableWriterHeaders(flags *pflag.FlagSet, sources []logs.LogSource) ([]st
 		maxContainerLen = max(maxArchLen, len(source.ContainerName))
 	}
 
-	if withTs {
+	if logsConfig.withTs {
```
victorchrollo14, line 666: think we should keep this as logsCfg, to keep it consistent.

PR #875, opened 2026-01-05
🎣 fix: display warning when cli config file parsing fails

### LINKED ISSUE #871: Display a warning to users on CLI config file parsing error
Recently we added an option to configure the `kubetail` CLI tool via a config file (https://github.com/kubetail-org/kubetail/pull/867). Currently, parsing errors are ignored silently. We should add a warning when a parsing error occurs to let users know about the problem.

smazmi:
Fixes #871

## Summary

This PR adds CLI config tests and makes some small improvements to the CLI config constructor and the logging config for the `serve` command.

## Changes

* Added `modules/shared/config/cli_test.go`
* Moved logging config to after flag parsing in `serve` sub-command
* Modifed code to re-use file extension on `modules/shared/config/cli.go`

## Submitter checklist

- [x] Rebase branch to HEAD
- [ ] Squash changes into one signed, single commit

### DIFF
--- a/modules/shared/config/cli.go
+++ b/modules/shared/config/cli.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 
 	"github.com/go-playground/validator/v10"
-	zlog "github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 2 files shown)


modules/shared/config/cli.go
```diff
@@ -89,21 +90,26 @@ func NewCLIConfigFromFile(configPath string) (*CLIConfig, error) {
 
 		// check extension
 		if len(filepath.Ext(configPath)) <= 1 {
-			return nil, fmt.Errorf("file %q must have a valid extension (e.g., .yaml, .json)", configPath)
+			zlog.Warn().Msgf("Config file %q must have a valid extension (e.g., .yaml, .json), using default configuration", configPath)
```
amorey: Using zlog without configuring the logger first will print messages like this:
```
$ ./bin/kubetail serve -c ~/.kubetail/config
{"level":"warn","time":"2026-01-06T08:24:49+03:00","message":"Config file \"home/.kubetail/config\" must have a valid extension (e.g., .yaml, .json), using default configuration"}
Started kubetail dashboard on http://localhost:7500/
```

To configure the logger you can look at this example:
https://github.com/kubetail-org/kubetail/blob/main/modules/cli/cmd/serve.go#L94-L99

After configuration the warning should look like this:
```
$ ./bin/kubetail serve -c ~/.kubetail/config
Warn: Config file "/home/.kubetail/config" must have a valid extension (e.g., .yaml, .json), using default configuration
Started kubetail dashboard on http://localhost:7500/
```

modules/shared/config/cli.go
```diff
@@ -89,21 +90,26 @@ func NewCLIConfigFromFile(configPath string) (*CLIConfig, error) {
...
+			zlog.Warn().Err(err).Msgf("Failed to read config file %q, using default configuration", configPath)
 		}
 	}
 
 	cfg := DefaultCLIConfig()
 
 	// unmarshal
 	if err := v.Unmarshal(cfg); err != nil {
-		return nil, err
+		zlog.Warn().Err(err).Msg("Failed to unmarshal config, using default configuration")
```
amorey: Let's leave this as a hard fail

smazmi:
> There were a lot of conflicting changes in `main` so I rebased the PR myself and at the same time changed the behavior so we return config parsing errors instead of just warning. I kept the CLI config test code but re-organized it as an example of how to write tests in the future.

Got it, Thanks for the review and merge!

PR #876, opened 2026-01-07
✨Split monolithic config into component-specific config files 

### LINKED ISSUE #873: Split monolithic config into component-specific config files
Currently the struct for the dashboard, cluster-api and cluster-agent configs is combined into one monolithic structure:
https://github.com/kubetail-org/kubetail/blob/main/modules/shared/config/config.go

The cluster-agent portion is already obsolete because that code has been moved to Rust and we also just added a new standalone CLI config too (https://github.com/kubetail-org/kubetail/blob/main/modules/shared/config/cli.go) so we're already moving towards component-specific configs instead of one monolith.

Now that we've already moved half-way we should finish the job and split out the dashboard and cluster-api configs too.

In order to do so we need to do the following:

* [ ] Create a component-specific config module for each Go component:
  * [ ] `modules/dashboard/pkg/config/config.go`
  * [ ] `modules/cluster-agent/pkg/config/config.go`
  * [ ] `modules/cli/pkg/config/config.go`
* [ ] Remove top-level parents from the Go config structs (e.g. `.Dashboard`, `.ClusterAPI`)
* [ ] Remove top-level parent from the Rust config struct (e.g. `.cluster_agent`)
* [ ] Remove `modules/shared/config` and update code pointing to it to use new modules
* [ ] Split the monolith seed file at [`hack/config.yaml`](https://github.com/kubetail-org/kubetail/blob/main/hack/config.yaml) into separate files
* [ ] Update the `ConfigMap`'s in the Tilt manifest [`hack/tilt/kubetail.yaml`](https://github.com/kubetail-org/kubetail/blob/main/hack/tilt/kubetail.yaml)
* [ ] Update the config generators in the helm templates:
  * [ ] [`kubetail.dashboard.config`](https://github.com/kubetail-org/helm-charts/blob/main/charts/kubetail/templates/_helpers.tpl#L133-L153)
  * [ ] [`kubetail.clusterAPI.config`](https://github.com/kubetail-org/helm-charts/blob/main/charts/kubetail/templates/_helpers.tpl#L278-L293)
  * [ ] [`kubetail.clusterAgent.config`](https://github.com/kubetail-org/helm-charts/blob/main/charts/kubetail/templates/_helpers.tpl#L410-L422)

This project is mostly in Go but there's a small Rust piece. It can be done by one person or a couple of people workign together.

DavidReque:
Fixes #873 

## Summary

Split the monolithic config structure into component-specific config modules, allowing each component (Dashboard, Cluster-API, CLI, Cluster-Agent) to have its own independent configuration without top-level parent keys.

## Changes

- Create component-specific config modules:
  - `modules/dashboard/pkg/config`
  - `modules/cluster-api/pkg/config`
  - `modules/cli/pkg/config`
- Remove top-level parents from config structs`(.Dashboard, .ClusterAPI, .cluster_agent)`
- Remove `modules/shared/config` and extract shared types to `modules/shared/configtypes`
- Split `hack/config.yaml` into separate seed files under `config/default/
`- Update `ConfigMaps `in `hack/tilt/kubetail.yaml`

## Submitter checklist

- [x] Rebase branch to HEAD


CLAUDE.md
```diff
@@ -64,7 +64,7 @@ pnpm graphql-codegen
 ```bash
 # Start dashboard server (requires config)
 cd modules/dashboard
-go run cmd/main.go -c hack/config.yaml
+go run cmd/main.go -c ../../config/default/dashboard.yaml
```
amorey: We can keep the dev config in `hack/config.yaml`
```suggestion
go run cmd/main.go -c hack/config.yaml
```

README.md
```diff
@@ -297,7 +297,7 @@ tilt up
 
 ```console
 cd modules/dashboard
-go run cmd/main.go -c hack/config.yaml
+go run cmd/main.go -c ../../config/default/dashboard.yaml
```
amorey:
```suggestion
go run cmd/main.go -c hack/config.yaml
```

modules/dashboard/pkg/config/config.go
```diff
@@ -0,0 +1,262 @@
...
+
+// Auth-mode
+type AuthMode = configtypes.AuthMode
+
+const (
+	AuthModeAuto  = configtypes.AuthModeAuto
+	AuthModeToken = configtypes.AuthModeToken
+)
+
+// Environment
```
amorey: What's the advantage of copying `configtypes.Environment` here? Why not use the shared type directly?

modules/shared/configtypes/configtypes.go
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package configtypes contains shared config-related types used across Go components.
+package configtypes
+
+// Auth-mode
```
amorey: What's the advantage of putting `AuthMode` in the shared module? Is it used outside of a dashboard context anywhere?

modules/dashboard/pkg/config/config.go
DavidReque: > What's the advantage of copying `configtypes.Environment` here? Why not use the shared type directly?

I did it this way so that consumers of the `dashboard/config` package only need one import. But you're right, it's cleaner to use the shared types directly. I'll refactor it.

PR #881, opened 2026-01-07
🐋 Add `--path` flag to `kubetail config init` command

### LINKED ISSUE #880: Add `--path` flag to `kubetail config init` command
Currently when you use the `kubetail config init` command, it initializes a default config file at the default path. We should add a `--path` flag that allows users to set their own path. The code is here:
https://github.com/kubetail-org/kubetail/blob/main/modules/cli/cmd/config_init.go

SimoHypers:
Fixes #880 

## Summary

This PR adds a `--path` flag to the `kubetail config init` command, allowing users to specify a custom path for the config file.

## Submitter checklist

- [x] Rebase branch to HEAD


modules/cli/cmd/config_init.go
```diff
@@ -56,5 +65,10 @@ var configInitCmd = &cobra.Command{
 }
 
 func init() {
+	configInitCmd.Flags().String(
+		"path",
+		"",
+		"custom path for the configuration file",
+	)
 	configCmd.AddCommand(configInitCmd)
```
amorey: Let's set it up like other flags in the CLI:
```suggestion
    configCmd.AddCommand(configInitCmd)
        
    flagset := configInitCmd.Flags()
    flagset.SortFlags = false
    flagset.String("path", "", "Target path for configuration file (default is $HOME/.kubetail/config.yaml")
```

modules/cli/cmd/config_init.go
```diff
@@ -27,9 +27,19 @@ var configInitCmd = &cobra.Command{
 		})
 	},
 	Run: func(cmd *cobra.Command, args []string) {
-		targetPath, err := config.DefaultConfigPath()
+		pathFlag, err := cmd.Flags().GetString("path")
```
amorey: I find it to be more readable when you keep the flag value variable stable:

```go
targetPath, _ := cmd.Flags().GetString("path")
if (targetPath == "") {
    tmp, err := config.DefaultConfigPath()
	if err != nil {
	    zlog.Fatal().Err(err).Msg("Unable to determine config path")
    }
    targetPath = tmp
}
```

PR #888, opened 2026-01-11
🐋 Add `--force` flag to `kubetail config init` command

### LINKED ISSUE #878: Add `--force` flag to `kubetail config init` command
Currently, if a config file already exists at the destination path the `config init` command exits with an error. We should add a `--force` flag that users can use to create a new file regardless of whether or not one exists there already. The code is here:
https://github.com/kubetail-org/kubetail/blob/main/modules/cli/cmd/config_init.go

smazmi:
Fixes #878

## Summary

This PR adds a `--force` flag to the `kubetail config init` command that allows users to overwrite existing configuration files. Previously, the command would exit with an error if a config file already existed at the target path.

## Changes

* Added `--force` boolean flag to the `config init` command
* Modified file existence check to allow overwriting when `--force` is true
* Added informational log message when overwriting: "Overwriting existing configuration file: <path>"

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/cli/cmd/config_init.go
+++ b/modules/cli/cmd/config_init.go
@@ -37,11 +37,10 @@ var configInitCmd = &cobra.Command{
 		}
 
 		force, _ := cmd.Flags().GetBool("force")
-		if _, err := os.Stat(targetPath); err == nil {
-			if !force {
+		if !force {
+			if _, err := os.Stat(targetPath); err == nil {
 				zlog.Fatal().Msgf("Configuration file already exists: %s", targetPath)
 			}
-			zlog.Info().Msgf("Overwriting existing configuration file: %s", targetPath)
 		}
 
 		configDir := filepath.Dir(targetPath)


modules/cli/cmd/config_init.go
```diff
@@ -36,8 +36,12 @@ var configInitCmd = &cobra.Command{
 			targetPath = tmp
 		}
 
+		force, _ := cmd.Flags().GetBool("force")
 		if _, err := os.Stat(targetPath); err == nil {
-			zlog.Fatal().Msgf("Configuration file already exists: %s", targetPath)
+			if !force {
+				zlog.Fatal().Msgf("Configuration file already exists: %s", targetPath)
+			}
+			zlog.Info().Msgf("Overwriting existing configuration file: %s", targetPath)
```
amorey: If the user used `--force` then I don't think we need the info message.
```suggestion
		force, _ := cmd.Flags().GetBool("force")
		
		if !force {
			if _, err := os.Stat(targetPath); err == nil {
				zlog.Fatal().Msgf("Configuration file already exists: %s", targetPath
			}
		}
```

smazmi: Understood, I'll commit the changes.

PR #931, opened 2026-01-25
🐋  Add checker to get upstream version

### LINKED ISSUE #925: Add upstream version checker
## Description

Implement a Go module that checks for the latest available versions of the Kubetail CLI and Helm chart from upstream GitHub repositories. The Dashboard will use this module to display "Upgrade Available" notifications to users when newer versions are available.

### Background

The "Upgrade Available" feature is split into three parts: frontend, backend, and interface. This issue covers the **backend** implementation.

#### CLI

The latest version of the CLI can be obtained from this endpoint:
https://api.github.com/repos/kubetail-org/kubetail/releases/latest

The “latest” tag is added by hand and the version returned from the endpoint is guaranteed to be the latest stable release.

#### Helm chart 

The latest version of the Helm chart can be obtained from this endpoint:
https://api.github.com/repos/kubetail-org/helm-charts/releases

The “latest” tag is applied to whichever version was published last so it’s not guaranteed to be the latest stable release (e.g. release candidates or patch releases). The module will need to parse the available versions to determine the actual latest stable version. 

### Deliverables

- [ ] A method to return the latest available CLI version from the upstream endpoint
- [ ] A method to return the latest available Helm chart version from the upstream endpoint

### Requirements

- [ ] Methods must return error responses to the caller

chilin0525:
Fixes #925 

## Summary

Add version checker to retrieve CLI and Helm chart versions from GitHub API with 12-hour caching.

## Changes

* Implement GitHub API integration to fetch latest CLI and Helm chart versions, with Helm chart selection based on `published_at` timestamp.
* Cache mechanism to prevent keep query Github API.

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/shared/versioncheck/github.go
+++ b/modules/shared/versioncheck/github.go
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
-	"strings"
 	"time"
 )
 
-var helmChartTagRegex = regexp.MustCompile(`^kubetail-(\d+\.\d+\.\d+)$`)
+var (
+	cliTagRegex       = regexp.MustCompile(`^cli/v(\d+\.\d+\.\d+)$`)
+	helmChartTagRegex = regexp.MustCompile(`^kubetail-(\d+\.\d+\.\d+)$`)
+)
 
 const (
 	cliReleasesURL        = "https://api.github.com/repos/kubetail-org/kubetail/releases/latest"
@@ -100,11 +102,11 @@ func (g *githubClient) fetchLatestCLIVersion(ctx context.Context) (string, error
 // parseCLITag extracts a semantic version from a CLI tag name.
 // Expected format: "cli/vX.Y.Z"
 func parseCLITag(tag string) (string, error) {
-	if !strings.HasPrefix(tag, "cli/v") {
+	matches := cliTagRegex.FindStringSubmatch(tag)
+	if len(matches) != 2 {
 		return "", fmt.Errorf("unexpected tag_name format: %s", tag)
 	}
-	version := strings.TrimPrefix(tag, "cli/v")
-	return version, nil
+	return matches[1], nil
 }
 
 func (g *githubClient) fetchLatestHelmChartVersion(ctx context.Context) (string, error) {
--- a/modules/shared/versioncheck/github_test.go
+++ b/modules/shared/versioncheck/github_test.go
@@ -179,6 +179,7 @@ func TestParseCLITag(t *testing.T) {
 		wantErr bool
 	}{
 		{"valid tag", "cli/v0.11.1", "0.11.1", false},
+		{"invalid tag", "cli/v", "", true},
 		{"missing cli prefix", "v0.11.1", "", true},
 		{"no prefix at all", "0.11.1", "", true},
 		{"empty string", "", "", true},


modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (c *checker) GetLatestVersions() *LatestVersions {
+	return &LatestVersions{
+		CLI:       c.GetLatestCLIVersion(),
+		HelmChart: c.GetLatestHelmChartVersion(),
+	}
+}
+
+func (c *checker) getLatestVersion(component Component, fetchFunc func(context.Context) (string, error)) (string, error) {
```
DavidReque: **Race condition**:The check-delete-fetch pattern is not thread-safe. If multiple goroutines call this when cache expires, they'll all pass the cache check and make separate GitHub requests.

Consider using `util.SyncGroup.LoadOrCompute()` which already exists in this codebase https://github.com/kubetail-org/kubetail/blob/main/modules/shared/util/syncgroup.go and handles this atomically.

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+			userAgent: "kubetail-version-checker",
+		},
+		cache:    util.SyncMap[Component, cacheEntry]{},
+		cacheTTL: DefaultCacheTTL,
+	}
+}
+
+func (c *checker) GetLatestCLIVersion() *VersionInfo {
+	info := &VersionInfo{
```
DavidReque: `LastChecked` is set before checking the cache, so if the value comes from cache, it doesn't reflect when it was actually fetched.

Consider storing `lastChecked` in the cache entry instead.

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,144 @@
...
+	expiration time.Time
+}
+
+type checker struct {
+	githubClient *githubClient
+	cache        util.SyncMap[Component, cacheEntry]
+	cacheTTL     time.Duration
+}
+
+func NewChecker() Checker {
```
DavidReque: The codebase uses the functional options pattern (see
[`modules/shared/helm/helm.go`](https://github.com/kubetail-org/kubetail/blob/4f259158a8a112809e3e5ed82c0ae45a69472a45/modules/shared/helm/helm.go#L381)). Consider adding options for testability and configurability:

```go
type CheckerOption func(*checker)

func WithCacheTTL(ttl time.Duration) CheckerOption { ... }
func WithHTTPClient(client *http.Client) CheckerOption { ... }

func NewChecker(options ...CheckerOption) Checker { ... }

File: modules/shared/versioncheck/github.go [hunk 4]
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("GitHub API returned status %d", resp.StatusCode)
+	}
+
+	var release GitHubRelease
```
DavidReque: `TagName` could be empty after successful JSON decode. Add validation:

```go
if release.TagName == "" {
    return "", fmt.Errorf("release tag_name is empty")
}
		continue
		}

		if latestRelease == nil || publishedAt.After(latestPublishedTime) {
			latestRelease = &r
			latestPublishedTime = publishedAt
		}
	}
```

modules/shared/versioncheck/github.go
```diff
@@ -0,0 +1,138 @@
...
+
+	var releases []GitHubRelease
+	if err := json.NewDecoder(resp.Body).Decode(&releases); err != nil {
+		return "", err
+	}
+
+	var latestRelease *GitHubRelease
+	var latestPublishedTime time.Time
+
+	for _, r := range releases {
```
DavidReque, line 141: The `helm-charts` repo might contain multiple charts. Consider filtering releases by the `kubetail-` prefix:

`if !strings.HasPrefix(r.TagName, "kubetail-") {
    continue
}`

modules/shared/versioncheck/github.go
```diff
@@ -0,0 +1,138 @@
...
+	"net/http"
+	"time"
+)
+
+const (
+	cliReleasesURL        = "https://api.github.com/repos/kubetail-org/kubetail/releases/latest"
+	helmChartsReleasesURL = "https://api.github.com/repos/kubetail-org/helm-charts/releases"
+)
+
+type GitHubRelease struct {
```
DavidReque: `GitHubRelease` is exported but `githubClient` is not. If this struct is only used internally, consider making it unexported (`githubRelease`).

modules/shared/versioncheck/github.go
chilin0525: Resolved in [84d30c7](https://github.com/kubetail-org/kubetail/pull/931/commits/84d30c73402bb9e38d86c5a66d037a036e55b091).

modules/shared/versioncheck/github.go
chilin0525: This was changed to unexported in [130c9d8](https://github.com/kubetail-org/kubetail/pull/931/commits/130c9d8a9ed2f5c5a3b3fbbd5a3d2f18672465ad).

modules/shared/versioncheck/versioncheck.go
chilin0525: Resolved in [2cb7bc4](https://github.com/kubetail-org/kubetail/pull/931/commits/2cb7bc40b2136247f59ac11e86711169616c09eb).

modules/shared/versioncheck/github.go
chilin0525: I've implemented the TagName validation as suggested in [1a3964a](https://github.com/kubetail-org/kubetail/pull/931/commits/1a3964acaa4c6009b325b01ce123e78e71dd9c6b), but wanted to confirm since the example mixed logic from both functions. Does this change match your intention?

modules/shared/versioncheck/versioncheck.go
chilin0525: Resolved in [cfc3650](https://github.com/kubetail-org/kubetail/pull/931/commits/cfc365095755393ce8a74ef8667d5a4525548a09). I also added `Load` and `Delete` to `SyncGroup` to support TTL-based cache invalidation.

modules/shared/versioncheck/versioncheck.go
chilin0525: Added functional options pattern to version checker in [6245ab6](https://github.com/kubetail-org/kubetail/pull/931/commits/6245ab6871d89546d20e41b2dce48be0e9113f5e).

DavidReque (approved):
Thanks for addressing all the feedback! The changes look great

I approve the changes 🎉

amorey:
@codex review

modules/shared/versioncheck/github.go
```diff
@@ -0,0 +1,151 @@
...
+	var release githubRelease
+	if err := json.NewDecoder(resp.Body).Decode(&release); err != nil {
+		return "", err
+	}
+
+	if release.TagName == "" {
+		return "", fmt.Errorf("release tag_name is empty")
+	}
+
+	return release.TagName, nil
```
amorey: `tag_name` has the format `cli/vX.Y.Z` so we need to remove the prefix to get the version number

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+func NewChecker(options ...CheckerOption) Checker {
+	c := &checker{
+		githubClient: &githubClient{
+			httpClient: &http.Client{
+				Timeout: defaultTimeout,
+			},
+			userAgent: "kubetail-version-checker",
```
amorey, line 54: A user agent with this format would be more conventional:
`kubetail/X.Y.Z (<component>; env=<env>; <os>; <arch>) <lang>/<lang-version>`

For example, this would be the value for a request coming from the dashboard running on my laptop:
`kubetail/0.11.0 (dashboard; env=desktop; Darwin; arm64) Go/1.24.7`

Would this be easy to implement or should we save it for a follow-on PR?

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,170 @@
...
+const (
+	DefaultCacheTTL = 12 * time.Hour
+	defaultTimeout  = 10 * time.Second
+)
+
+type Component string
+
+const (
+	ComponentCLI       Component = "cli"
+	ComponentHelmChart Component = "helm-chart"
```
amorey: Should these constants be exported?

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,170 @@
...
+	info.LastChecked = lastChecked
+	if err != nil {
+		zlog.Debug().Err(err).Msg("Failed to get latest Helm chart version")
+		info.Error = err
+		return info
+	}
+
+	info.Version = version
+	return info
+}
```
amorey: A couple of points:
* I would expect these functions to passthrough errors to the caller in the function response `(*VersionInfo, error)`
* If we modify the functions to passthrough errors then I don't think they need to also log the errors

modules/shared/versioncheck/versioncheck.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (c *checker) GetLatestVersions() *LatestVersions {
+	return &LatestVersions{
+		CLI:       c.GetLatestCLIVersion(),
+		HelmChart: c.GetLatestHelmChartVersion(),
+	}
+}
```
amorey: Are we sure we need a function that returns both versions simultaneously? Adding this now might be premature.

modules/shared/versioncheck/github.go
```diff
@@ -0,0 +1,151 @@
...
+			latestRelease = &r
+			latestPublishedTime = publishedAt
+		}
+	}
+
+	if latestRelease == nil {
+		return "", fmt.Errorf("no valid releases found")
+	}
+
+	return latestRelease.TagName, nil
```
amorey: We need to strip the prefix (`kubetail-`) to get the version number. Also, it doesn't look like we're excluding non-stable releases which can be identified by a hyphen after the version number (e.g. `kubetail-0.17.0-rc1`).

chilin0525: The parsing issue has been resolved in https://github.com/kubetail-org/kubetail/pull/931/changes/98b93ab99f06fb77da49039f5d07d56cae56f866

modules/shared/versioncheck/github.go
chilin0525: The parsing issue has been resolved in https://github.com/kubetail-org/kubetail/pull/931/changes/98b93ab99f06fb77da49039f5d07d56cae56f866

modules/shared/versioncheck/versioncheck.go
chilin0525: Removed in https://github.com/kubetail-org/kubetail/pull/931/changes/3628cdffd25dfc8d937ca62475b0fa2a253ebf84

amorey:
@chilin0525 The list looks good. Just one small suggestion - we may want to strip the "v" in the version number too so we can do semver comparisons more easily later.

modules/shared/versioncheck/versioncheck.go
chilin0525: * Error passthrough to the caller has been implemented in:  https://github.com/kubetail-org/kubetail/pull/931/changes/a8f202db593dbca5fa8d697a03ff28b02a39cdfd
* Logging has already been removed in:  https://github.com/kubetail-org/kubetail/pull/931/changes/3628cdffd25dfc8d937ca62475b0fa2a253ebf84

chilin0525:
@amorey Do you mean removing the version prefix for the CLI, like `cli/v0.11.1` → `0.11.1`? If so, that's already the current implementation.

modules/shared/versioncheck/versioncheck.go
chilin0525: These constants have already been removed in: https://github.com/kubetail-org/kubetail/pull/931/changes/c709444e18d2d2104be597b8ec5369eeff0f18b5

modules/shared/versioncheck/github.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		publishedAt, err := time.Parse(time.RFC3339, r.PublishedAt)
+		if err != nil {
+			continue
+		}
+
+		if latestRelease == nil || publishedAt.After(latestPublishedTime) {
+			latestRelease = &r
+			latestPublishedTime = publishedAt
```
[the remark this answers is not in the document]
chilin0525: Resolved in https://github.com/kubetail-org/kubetail/pull/931/changes/98b93ab99f06fb77da49039f5d07d56cae56f866.

DavidReque (approved):

amorey:
Merged! 🥳

PR #934, opened 2026-01-27
🐋 Implement K8s authentication caching

### LINKED ISSUE #631: Implement caching during k8s authentication
# Overview

Currently, when the cluster agent receives a gRPC request, it authenticates the request by:
- Creating a new k8s client.
- Doing a request to the k8s API to authenticate the request.

This flow has a performance penalty as we authenticate on each request.

# Suggested solution

We should implement caching to reduce the number of requests. This means that we should keep in memory for each token the verbs and namespaces that it gives access to.

joaquinalmora:
Fixes #631

## Summary
Implement Kubernetes authentication result caching to reduce load on the Kubernetes API server during authorization checks.

## Changes
- Implement process-scoped authorization cache with 10,000 max entries and 5-minute TTL
- Cache key: (SHA256(token), namespace, verb) tuple
- Add dependencies: moka, sha2, secrecy
- Update `Authorizer::new()` to accept shared cache
- Modify all service handlers to pass cache to authorizer
- Store authorization results in cache after K8s API responses

## Verification
- All tests pass (18/18 in cluster_agent crate)
- cargo fmt, cargo clippy, cargo test verified
- Updated test fixtures to include auth cache

### DIFF
--- a/crates/cluster_agent/src/log_metadata.rs
+++ b/crates/cluster_agent/src/log_metadata.rs
@@ -19,7 +19,7 @@ use std::env;
 use std::fs::File;
 use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, LazyLock};
+use std::sync::LazyLock;
 use tokio_util::sync::CancellationToken;
 use tracing::debug;
 
--- a/crates/cluster_agent/src/log_records.rs
+++ b/crates/cluster_agent/src/log_records.rs
@@ -14,7 +14,6 @@
 
 use std::fs;
 use std::path::PathBuf;
-use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
 use tokio::sync::mpsc::{self};
--- a/crates/cluster_agent/src/main.rs
+++ b/crates/cluster_agent/src/main.rs
@@ -16,7 +16,6 @@ use std::error::Error;
 use std::fs::read_to_string;
 use std::path::PathBuf;
 use std::str::FromStr;
-use std::sync::Arc;
 
 use clap::{ArgAction, arg, command, value_parser};
 use tokio::signal::ctrl_c;


crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,17 +23,50 @@ use kube::Config;
...
+    namespace: String,
+    verb: String,
+}
+
+/// Value for the authorization cache: allowed (true/false)
+pub type CacheValue = bool;
+
+/// Creates a process-scoped authorization cache
+pub fn create_auth_cache() -> Arc<Cache<CacheKey, CacheValue>> {
+    Arc::new(
```
amorey: Is the outer `Arc` necessary? From the docs it looks like `Cache` can be cloned directly.

crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,17 +23,50 @@ use kube::Config;
...
+        Cache::builder()
+            .max_capacity(10_000)
+            .time_to_live(Duration::from_secs(300))
```
amorey: Just a note, we should make this configurable in the future.

crates/cluster_agent/src/log_metadata.rs
```diff
@@ -51,15 +52,22 @@ pub struct LogMetadataImpl {
     task_tracker: TaskTracker,
     logs_dir: PathBuf,
     node_name: String,
+    auth_cache: Arc<Cache<crate::authorizer::CacheKey, crate::authorizer::CacheValue>>,
```
amorey: We can define the cache type in `authorizer.rs` and import it from there so it's defined in one place

crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,17 +23,50 @@ use kube::Config;
...
 
 /// Checks that the the k8s doing the request has proper rights to access the log files.
 #[cfg(not(test))]
 impl Authorizer {
     /// Creates a new Authorizer, using the k8s authorization token to construct the proper
     /// client set during authorization.
-    pub async fn new(request_metadata: &MetadataMap) -> Result<Self, Status> {
+    pub async fn new(
+        request_metadata: &MetadataMap,
+        auth_cache: Arc<Cache<CacheKey, CacheValue>>,
```
amorey: Instead of initializing the `Authorizer` once per request and passing in a shared `auth_cache` did you consider refactoring the `Authorizer` so that it can be shared acrosss requests itself?

amorey:
>     * Regarding the shared authorizer, the existing design stores the user's token in `k8s_config` so each request needs its own Authorizer instance (different users = different tokens). I could move the token extraction into `is_authorized` so the Authorizer itself can be shared, should I do this as a follow-up or in this PR

I think it would be great to do it in this PR. A shared `Authorizer` would make the intention more apparent and make the code easier to read.

crates/cluster_agent/src/log_records.rs
```diff
@@ -112,12 +122,14 @@ impl LogRecordsService for LogRecordsImpl {
...
     ) -> Result<Response<Self::StreamForwardStream>, Status> {
-        let authorizer = Authorizer::new(request.metadata()).await?;
+        let request_metadata = request.metadata().clone();
         let request = request.into_inner();
         let file_path = self.get_log_filename(&request)?;
 
         let namespaces = vec![request.namespace.clone()];
-        authorizer.is_authorized(&namespaces, "list").await?;
+        self.authorizer
+            .is_authorized(&request_metadata, &namespaces, "get")
```
amorey: Let's move this to the other PR so one is not dependent on the other:
```suggestion
            .is_authorized(&request_metadata, &namespaces, "list")
```

crates/cluster_agent/src/log_records.rs
```diff
@@ -79,14 +87,16 @@ impl LogRecordsService for LogRecordsImpl {
...
+        let request_metadata = request.metadata().clone();
         let request = request.into_inner();
         let file_path = self.get_log_filename(&request)?;
         let (tx, rx) = mpsc::channel(100);
         let local_ctx = self.ctx.child_token();
 
         let namespaces = vec![request.namespace.clone()];
-        authorizer.is_authorized(&namespaces, "list").await?;
+        self.authorizer
+            .is_authorized(&request_metadata, &namespaces, "get")
```
amorey:
```suggestion
            .is_authorized(&request_metadata, &namespaces, "list")
```

crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,67 +23,133 @@ use kube::Config;
...
+/// Creates a process-scoped authorization cache
+pub fn create_auth_cache() -> AuthCache {
+    Cache::builder()
+        .max_capacity(10_000)
+        .time_to_live(Duration::from_secs(300))
+        .build()
+}
+
 #[allow(dead_code)]
+#[derive(Debug)]
```
amorey: I haven't verified it yet, but we should be able to derive the Clone trait here then use `.clone()` to share ownership instead of wrapping this in `Arc`:
```suggestion
#[derive(Clone, Debug)]
```

crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,67 +23,133 @@ use kube::Config;
...
 
 /// Checks that the the k8s doing the request has proper rights to access the log files.
 #[cfg(not(test))]
 impl Authorizer {
-    /// Creates a new Authorizer, using the k8s authorization token to construct the proper
-    /// client set during authorization.
-    pub async fn new(request_metadata: &MetadataMap) -> Result<Self, Status> {
-        let token = request_metadata
+    /// Creates a new Authorizer that can be shared across requests.
+    pub async fn new(auth_cache: AuthCache) -> Result<Self, Status> {
```
amorey: It would be more ergonomic to make `auth_cache` internal to `Authorizer` so the caller doesn't have to initialize it.

crates/cluster_agent/src/authorizer.rs
```diff
@@ -23,67 +23,133 @@ use kube::Config;
...
+pub type CacheValue = bool;
+
+/// Type alias for the authorization cache
+pub type AuthCache = Cache<CacheKey, CacheValue>;
+
+/// Creates a process-scoped authorization cache
+pub fn create_auth_cache() -> AuthCache {
+    Cache::builder()
+        .max_capacity(10_000)
+        .time_to_live(Duration::from_secs(300))
```
amorey: Let's decrease the TTL so that it helps with bursts but doesn't give users access beyond what the permissions intend.
```suggestion
        .time_to_live(Duration::from_secs(30))
```

amorey:
Hmm the DCO check still isn't passing... can you try squashing the commits with a sign-off?

PR #966, opened 2026-02-15
cli/logs: persist dot color assignments across runs

### LINKED ISSUE #683: Create plan for saving "color dot" state across `kubetail logs` instantiations
Our goal is to modify the default output of the kubetail logs command to look like this (see https://github.com/kubetail-org/kubetail/issues/680 for more context):
Image

After the "color dot" identifer is implemented, we need to add a way to save state across instantiations so that the containers get assigned the same color anytime the tool is used.

Before implementing this feature we need to consider these questions:
- How should we save state (text file, sqlite)?
- How should we deal with concurrency issues?
- How should we handle schema changes on upgrade?
- How should we handle cleanup for dead containers?

After answering these questions let's determine the best way to implement the feature.

amirHdev:
Closes #683

## Summary
Persist `kubetail logs` dot color assignments so containers keep the same color across CLI invocations.

## Changes
- Added a persistent dot color state manager:
  - file-backed JSON store at `~/.kubetail/state/log-dot-colors.v1.json`
  - schema versioning (`schemaVersion`) for upgrade handling
  - lock-based write coordination for concurrent CLI processes
  - atomic state writes
- Wired persistent color assignment into `logs` rendering flow.
- Kept safe fallback behavior: if persistence fails, dot color still renders via hash-based assignment.
- Added stale entry cleanup with TTL .
- Added tests for:
  - persistence across instantiations
  - schema mismatch/reset handling
  - stale entry cleanup

## Validation
- `cd modules/cli && go test ./cmd/...`

### DIFF
--- a/modules/cli/cmd/logs.go
+++ b/modules/cli/cmd/logs.go
@@ -465,6 +465,21 @@ func printLogs(rootCtx context.Context, cmd *cobra.Command, cmdCfg *logsCmdConfi
 		writer.Flush()
 	}
 
+	var dotAssigner *dotColorAssigner
+	if cmdCfg.withDot {
+		assigner, err := newDotColorAssigner()
+		if err != nil {
+			fmt.Fprintf(cmd.OutOrStderr(), "warning: dot color persistence disabled: %v\n", err)
+		} else {
+			dotAssigner = assigner
+			defer func() {
+				if err := dotAssigner.Close(); err != nil {
+					fmt.Fprintf(cmd.OutOrStderr(), "warning: dot color persistence disabled: %v\n", err)
+				}
+			}()
+		}
+	}
+
 	// Write rows
 	var firstRecord, lastRecord *logs.LogRecord
 	for record := range stream.Records() {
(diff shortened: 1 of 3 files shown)


amorey:
@amirHdev Thanks again for the PR and apologies for the delay. There are definitely some useful things in this PR but I think we should hold off on implementing the feature until we get more feedback from users. The current hashing solution seems to be "good enough" so I'm hesitant to make big changes until users ask for it. However, this PR is useful for the planning stage so let's keep it open to help us later.

Here's another issue you might find interesting to work on: https://github.com/kubetail-org/kubetail/issues/970.

Currently, users are required to modify columns individually in the CLI output (e.g. `--hide-ts`, `--with-region`, `--with-zone`) which can get unwieldy. Instead I want to consolidate the column manipulation into these flags:

```
--columns
--add-columns
--remove-columns
```

After implementing that we should also make it configurable via the local config file. If you're interested in working on this let me know!

amirHdev:
> Sounds great! Starting with a proposal sounds like a good place to start. I’m happy to answer any questions you have technical or otherwise. You can DM me on discord if you want to chat too.

how can i connect with you in the Discord?

themavik (commented):
Persisting color assignments across runs is a nice UX win. Quick thought: if the stored color map grows unbounded over time (many transient pod names), is there a cap or eviction strategy?

PR #967, opened 2026-02-19
🐋 Implement backend API for Dashboard version status

### LINKED ISSUE #926: Implement backend API for Dashboard version status
## Description

Implement a backend API endpoint that the Dashboard frontend can query to determine version status. The endpoint will compare the current app version against the latest upstream version (using #925) and return data the frontend needs to display "Upgrade Available" notifications.

### Background

The "Upgrade Available" feature is split into three parts: frontend, backend, and interface. This issue covers the **interface** implementation.

#### Desktop

In `DesktopMode`, the Dashboard is started using the Kubetail CLI tool so the frontend must be aware of newer version of the CLI available. In addition, users can switch between clusters in `DesktopMode` so the frontend must be aware of newer versions available for all available cluster contexts. We will use the Helm chart version for cluster version checks.

#### Cluster

In `ClusterMode`, the Dashboard runs from the `kubetail-dashboard` Docker image and is typically deployed using Helm so the frontend only needs to be aware of newer version of the Helm chart available.

### Deliverables

// TODO

### Requirements

- [ ] If a cluster helm version is unavailable, ignore version checks for that particular cluster

chilin0525:
Fixes #926 

## Summary

<-- Explain the goal of your PR  -->

## Changes

<-- Explain the changes in your PR -->

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/dashboard/graph/schema.resolvers.go
+++ b/modules/dashboard/graph/schema.resolvers.go
@@ -19,7 +19,6 @@ import (
 	"github.com/kubetail-org/kubetail/modules/shared/helm"
 	"github.com/kubetail-org/kubetail/modules/shared/k8shelpers"
 	"github.com/kubetail-org/kubetail/modules/shared/logs"
-	"github.com/kubetail-org/kubetail/modules/shared/versioncheck"
 	zlog "github.com/rs/zerolog/log"
 	"helm.sh/helm/v3/pkg/release"
 	appsv1 "k8s.io/api/apps/v1"
@@ -31,18 +30,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/utils/ptr"
 )
-
-var newVersionChecker = func() versioncheck.Checker {
-	return versioncheck.NewChecker()
-}
-
-type helmListClient interface {
-	ListReleases() ([]*release.Release, error)
-}
-
-var newHelmListClient = func(opts ...helm.ClientOption) helmListClient {
-	return helm.NewClient(opts...)
-}
 
 // Object is the resolver for the object field.
 func (r *appsV1DaemonSetsWatchEventResolver) Object(ctx context.Context, obj *watch.Event) (*appsv1.DaemonSet, error) {
@@ -570,8 +557,7 @@ func (r *queryResolver) CliVersionStatus(ctx context.Context) (*model.VersionSta
 		return nil, nil
 	}
 
-	vc := newVersionChecker()
-	latestInfo, err := vc.GetLatestCLIVersion()
+	latestInfo, err := r.versionChecker.GetLatestCLIVersion()
 	if err != nil {
 		zlog.Warn().Err(err).Msg("Failed to check latest CLI version")
 		return nil, nil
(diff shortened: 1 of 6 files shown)


modules/dashboard/graph/server.go
```diff
@@ -51,11 +52,16 @@ func NewServer(cfg *config.Config, cm k8shelpers.ConnectionManager) *Server {
...
 
+	// Init version checker
+	vc := versioncheck.NewChecker()
+
 	// Init resolver
 	r := &Resolver{
 		cfg:               cfg,
 		cm:                cm,
 		hm:                hm,
+		versionChecker:    vc,
```
amorey: Instead of attaching an instance of the version checker to the resolver can we just instantiate it on the fly when the client makes a call to the checker endpoint?

modules/dashboard/graph/server.go
```diff
@@ -51,11 +52,16 @@ func NewServer(cfg *config.Config, cm k8shelpers.ConnectionManager) *Server {
...
+		cliVersion:        cfg.CLIVersion,
```
amorey: You can access `CLIVersion` through `cfg` so it shouldn't be necessary to attach it to the resolver.

modules/dashboard/graph/server.go
chilin0525: Thanks for review! Resolved in 3fae118d8d01e343ea19da6c0a2331648996ccda.

modules/dashboard/graph/server.go
chilin0525: Thanks for review! Resolved in 3fae118d8d01e343ea19da6c0a2331648996ccda.

modules/dashboard/graph/schema.resolvers.go
```diff
@@ -29,6 +32,18 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+var newVersionChecker = func() versioncheck.Checker {
```
amorey: Now I see why you had attached the version checker to the resolver object previously. I think for dependency injection, it's more idiomatic to do it the way you had it before. Apologies for the confusion.

modules/dashboard/graph/schema.resolvers.go
```diff
@@ -29,6 +32,18 @@ import (
...
+	return versioncheck.NewChecker()
+}
+
+type helmListClient interface {
+	ListReleases() ([]*release.Release, error)
+}
+
+var newHelmListClient = func(opts ...helm.ClientOption) helmListClient {
```
amorey: It looks like this might belong on the resolver object for dependency injection too. You can put the helper interface in `schema.helpers.go`.

modules/dashboard/graph/schema.resolvers.go
chilin0525: No worries at all and thanks for clarifying! I’ll restore the previous approach.

chilin0525: Resolved in 7e4b6f3509787d4505f2e1a9fb53c5fa40206141.

modules/dashboard/graph/schema.resolvers.go
chilin0525: Resolved in 7e4b6f3509787d4505f2e1a9fb53c5fa40206141.

amorey (approved):

PR #971, opened 2026-02-21
🐋 Add unified column configuration flags and config support

### LINKED ISSUE #970: Change `kubetail logs` column handling
Currently `kubetail logs` enables column configuration via individual flags (e.g. `--with-zone`). We should allow users to configure columns using these flags instead:
```
--columns
--add-columns
--remove-columns
```

amirHdev:
Closes #970

## Summary
This PR introduces a unified column configuration model for `kubetail logs` using:
- `--columns`
- `--add-columns`
- `--remove-columns`

It also adds equivalent local config support so users can define defaults in their CLI config file.

## Changes
- Added new CLI flags for logs column selection:
  - `--columns` (replace column set)
  - `--add-columns` (enable additional columns)
  - `--remove-columns` (disable columns)
- Added config keys under `commands.logs`:
  - `columns`
  - `add-columns`
  - `remove-columns`
- Implemented column resolution with precedence and validation:
  - built-in defaults
  - config overrides
  - CLI overrides
  - backward-compatible handling of existing `--with-*` / `--hide-*` flags
- Updated default config templates:
  - `modules/cli/assets/config.yaml`
  - `modules/cli/assets/config.toml`
  - `modules/cli/assets/config.json`
- Added/updated tests for:
  - config parsing
  - flag behavior
  - precedence
  - invalid column handling

## Test
- `cd modules/cli && go test ./cmd/... ./pkg/config/...`

### DIFF
--- a/modules/cli/assets/config.yaml
+++ b/modules/cli/assets/config.yaml
@@ -58,6 +58,17 @@ commands:
     #
     kube-context: ""
 
+    ## columns ##
+    #
+    # Full set of output columns for 'logs' records.
+    # Allowed values: timestamp,dot,node,region,zone,os,arch,namespace,pod,container
+    #
+    # Default value: ["timestamp", "dot"]
+    #
+    columns:
+      - timestamp
+      - dot
+
   ## serve ##
   #
   # Settings for the dashboard server
--- a/modules/cli/cmd/logs.go
+++ b/modules/cli/cmd/logs.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"slices"
 	"strings"
 	"syscall"
 	"text/template"
@@ -75,23 +76,10 @@ type logsCmdConfig struct {
 	osList     []string
 	archList   []string
 	nodeList   []string
+	columns    []string
 
-	hideHeader bool
-	hideTs     bool
-	hideDot    bool
-
-	withTs        bool
-	withDot       bool
+	hideHeader    bool
 	allContainers bool
-
-	withNode      bool
-	withRegion    bool
-	withZone      bool
-	withOS        bool
-	withArch      bool
-	withNamespace bool
-	withPod       bool
-	withContainer bool
 	withCursors   bool
 
 	raw bool
@@ -254,6 +242,10 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
 
 	v.BindPFlag("general.kubeconfig", flags.Lookup(KubeconfigFlag))
 	v.BindPFlag("commands.logs.kube-context", flags.Lookup(KubeContextFlag))
+	if flags.Changed("columns") {
+		columns, _ := flags.GetStringSlice("columns")
+		v.Set("commands.logs.columns", columns)
+	}
 
 	if headFlag.IsValueProvided {
 		v.Set("commands.logs.head", headFlag.Value)
@@ -292,36 +284,19 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
 	nodeList, _ := flags.GetStringSlice("node")
 
 	hideHeader, _ := flags.GetBool("hide-header")
-	hideTs, _ := flags.GetBool("hide-ts")
-	hideDot, _ := flags.GetBool("hide-dot")
-
-	withTs := !hideTs
-	withDot := !hideDot
 	allContainers, _ := flags.GetBool("all-containers")
 
-	withNode, _ := flags.GetBool("with-node")
-	withRegion, _ := flags.GetBool("with-region")
-	withZone, _ := flags.GetBool("with-zone")
-	withOS, _ := flags.GetBool("with-os")
-	withArch, _ := flags.GetBool("with-arch")
-	withNamespace, _ := flags.GetBool("with-namespace")
-	withPod, _ := flags.GetBool("with-pod")
-	withContainer, _ := flags.GetBool("with-container")
 	withCursors, _ := flags.GetBool("with-cursors")
+
+	columns := resolveLogsColumns(cfg)
+	addColumns, _ := flags.GetStringSlice("add-columns")
+	removeColumns, _ := flags.GetStringSlice("remove-columns")
+	columns = applyColumnAddRemove(columns, addColumns, removeColumns)
 
 	raw, _ := flags.GetBool("raw")
 	if raw {
 		hideHeader = true
-		withTs = false
-		withNode = false
-		withRegion = false
-		withZone = false
-		withOS = false
-		withArch = false
-		withNamespace = false
-		withPod = false
-		withContainer = false
-		withDot = false
+		columns = []string{}
 		allContainers = false
 	}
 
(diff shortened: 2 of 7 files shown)


modules/cli/cmd/logs.go
```diff
@@ -450,6 +459,139 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
 	return cmdCfg, nil
 }
 
+func resolveLogsColumns(flags *pflag.FlagSet, cfg *config.Config) (bool, bool, bool, bool, bool, bool, bool, bool, bool, bool, error) {
```
amorey: Instead of returning a bool for each column it would simplify the code if we returned a single list of strings representing the resolved columns. That would also give the user control over the ordering of the columns.

modules/cli/cmd/logs.go
```diff
@@ -292,23 +306,18 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
-	withContainer, _ := flags.GetBool("with-container")
 	withCursors, _ := flags.GetBool("with-cursors")
 
+	withTs, withDot, withNode, withRegion, withZone, withOS, withArch, withNamespace, withPod, withContainer, err := resolveLogsColumns(flags, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	hideTs := !withTs
+	hideDot := !withDot
```
amorey: I don't think we need `hideTs` and `hideDot` anymore.

modules/cli/cmd/logs.go
```diff
@@ -450,6 +459,139 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
+	applyLegacyColumnBool(flags, "hide-ts", current, "ts", true)
+	applyLegacyColumnBool(flags, "hide-dot", current, "dot", true)
+	applyLegacyColumnBool(flags, "with-node", current, "node", false)
+	applyLegacyColumnBool(flags, "with-region", current, "region", false)
+	applyLegacyColumnBool(flags, "with-zone", current, "zone", false)
+	applyLegacyColumnBool(flags, "with-os", current, "os", false)
+	applyLegacyColumnBool(flags, "with-arch", current, "arch", false)
+	applyLegacyColumnBool(flags, "with-namespace", current, "namespace", false)
+	applyLegacyColumnBool(flags, "with-pod", current, "pod", false)
+	applyLegacyColumnBool(flags, "with-container", current, "container", false)
```
amorey: I don't think it's worth the added complexity to keep supporting the old flags and the CLI is pre-v1 so it's ok for this to be a breaking change.

amorey (changes requested):
Thanks for the quick turnaround! I added some suggestions inline. Overall, I think that if we return the resolved columns as a list and remove support for the legacy column flags it would simplify the code and make the feature easier to maintain.

amirHdev:
Thanks for the suggestions @amorey 

- Refactored column resolution to return an ordered []string (preserves user-defined column order).
- Removed legacy --with-* / --hide-* flags and compatibility logic.
- Removed hideTs/hideDot state.
- Simplified header/row rendering to iterate directly over resolved columns.

I also updated tests accordingly.

modules/cli/assets/config.yaml
```diff
@@ -58,6 +58,34 @@ commands:
     #
     kube-context: ""
 
+    ## columns ##
+    #
+    # Full set of output columns for 'logs' records.
+    # Allowed values: ts,dot,node,region,zone,os,arch,namespace,pod,container
+    # When empty, built-in defaults are used.
+    #
+    # Default value: []
```
amorey: Default value should be `["ts", "dot"]`

modules/cli/assets/config.yaml
```diff
@@ -58,6 +58,34 @@ commands:
...
+    #
+    columns: []
```
amorey:
```suggestion
    columns:
      - ts
      - dot
```

modules/cli/pkg/config/config.go
```diff
@@ -66,6 +69,9 @@ func DefaultCLIConfig() *Config {
 	cfg.Commands.Logs.KubeContext = ""
 	cfg.Commands.Logs.Head = 10
 	cfg.Commands.Logs.Tail = 10
+	cfg.Commands.Logs.Columns = []string{}
```
amorey:
```suggestion
	cfg.Commands.Logs.Columns = []string{"ts", "dot"}
```

modules/cli/cmd/logs.go
```diff
@@ -43,6 +44,21 @@ import (
...
+	"region",
+	"zone",
+	"os",
+	"arch",
+	"namespace",
+	"pod",
+	"container",
+}
+
+var defaultLogColumns = []string{"ts", "dot"}
```
amorey: It would be better to set the default in `DefaultConfig()` to let `NewConfig()` handle the file parsing and flag handling.

modules/cli/cmd/logs.go
```diff
@@ -450,6 +423,113 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
 	return cmdCfg, nil
 }
 
+func resolveLogsColumns(flags *pflag.FlagSet, cfg *config.Config) ([]string, error) {
```
amorey: `NewConfig()` already handles defaults, file parsing and flag overrides so you only need to resolve the columns in the `cfg` instance.

modules/cli/cmd/logs.go
```diff
@@ -450,6 +423,113 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
+		}
+	}
+	for _, col := range removeList {
+		updated = removeColumn(updated, col)
+	}
+
+	return updated, nil
+}
+
+func normalizeColumns(columns []string) ([]string, error) {
```
amorey: Trimming space and converting to lowercase is helpful but I don't think we need to check against a supported list of columns for now. Let's see what kind of mistakes users make before we decide if spelling errors should be a hard fail, warning or ignore.

modules/cli/cmd/logs.go
```diff
@@ -450,6 +423,113 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
+		return nil, fmt.Errorf("add-columns: %w", err)
+	}
+	removeList, err := normalizeColumns(removeColumns)
+	if err != nil {
+		return nil, fmt.Errorf("remove-columns: %w", err)
+	}
+
+	updated := append([]string{}, current...)
+	for _, col := range addList {
+		if !containsColumn(updated, col) {
```
amorey: You can use `slices.Contains()`

modules/cli/cmd/logs.go
```diff
@@ -450,6 +423,113 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
+func containsColumn(columns []string, column string) bool {
+	for _, col := range columns {
+		if col == column {
+			return true
+		}
+	}
+	return false
+}
+
+func removeColumn(columns []string, column string) []string {
```
amorey: Can be simplified with `slices.DeleteFunc()`

modules/cli/cmd/logs.go
```diff
@@ -475,38 +555,29 @@ func printLogs(rootCtx context.Context, cmd *cobra.Command, cmdCfg *logsCmdConfi
...
-		}
-		if cmdCfg.withNamespace {
-			row = append(row, orDefault(record.Source.Namespace, "-"))
-		}
-		if cmdCfg.withPod {
-			row = append(row, orDefault(record.Source.PodName, "-"))
-		}
-		if cmdCfg.withContainer {
-			row = append(row, orDefault(record.Source.ContainerName, "-"))
+		for _, col := range cmdCfg.columns {
```
amorey, line 481: 🔥

modules/cli/cmd/logs.go
```diff
@@ -450,6 +417,50 @@ func loadLogsCmdConfig(cmd *cobra.Command) (*logsCmdConfig, error) {
...
+
+	return updated
+}
+
+func normalizeColumns(columns []string) []string {
+	out := []string{}
+	seen := map[string]struct{}{}
+
+	for _, col := range columns {
+		for _, item := range strings.Split(col, ",") {
```
amorey: The split isn't necessary because the columns are already a list of strings.

modules/cli/pkg/config/config.go
```diff
@@ -41,9 +41,12 @@ type Config struct {
...
 		Logs struct {
-			KubeContext string `mapstructure:"kube-context"`
-			Head        int64  `mapstructure:"head"`
-			Tail        int64  `mapstructure:"tail"`
+			KubeContext   string   `mapstructure:"kube-context"`
+			Head          int64    `mapstructure:"head"`
+			Tail          int64    `mapstructure:"tail"`
+			Columns       []string `mapstructure:"columns"`
+			AddColumns    []string `mapstructure:"add-columns"`
+			RemoveColumns []string `mapstructure:"remove-columns"`
```
amorey: After thinking about it some more, I think we should remove `AddColumns` and `RemoveColumns` from the config file options. If a user wants to configure their columns it's more intutive to set the default in the settings file (using `Columns`) and then override them if necessary using the `--columns`, `--add-columns` and `--remove-columns` CLI flags.

modules/cli/cmd/logs.go
```diff
@@ -663,48 +665,39 @@ func getTableWriterHeaders(cmdCfg *logsCmdConfig, sources []logs.LogSource) ([]s
...
-	if cmdCfg.withPod {
-		headers = append(headers, "POD")
-		colWidths = append(colWidths, maxPodLen)
-	}
-	if cmdCfg.withContainer {
-		headers = append(headers, "CONTAINER")
-		colWidths = append(colWidths, maxContainerLen)
+	for _, col := range cmdCfg.columns {
+		switch col {
+		case "ts":
```
amorey: Let's rename "ts" to "timestamp". Previously we were using `ts` as a shorthand for the flagname but it's more intutive for each column name to correspond to its header label (except for "dot").

modules/cli/cmd/logs_test.go
```diff
@@ -78,6 +70,74 @@ func TestLoadLogConfig(t *testing.T) {
...
+		cmd := &cobra.Command{}
+		addLogsCmdFlags(cmd)
+		cmd.Flags().String("config", "", "")
+		assert.NoError(t, cmd.Flags().Set("config", cfgPath))
+
+		cmdCfg, err := loadLogsCmdConfig(cmd)
+		assert.NoError(t, err)
+
+		assert.Equal(t, []string{"namespace"}, cmdCfg.columns)
+	})
```
amorey: Config tests should go in `pkg/config/config_test.go` (though I think we can remove this partiular test altogether).

amorey (changes requested):
Looking good! I added one comment about a test that doensn't belong in `logs_test.go`, otherwise tests look good. Other than that, I think we should rename "ts" -> "timestamp" and remove `AddColumns` and `RemoveColumns` from the CLI config options.

PR #998, opened 2026-03-04
🐋 Add upgrade notification banner and notifications integration

### LINKED ISSUE #927: Add "Upgrade Available" banner and notifications integration
## Description

Implement an "Upgrade Available" banner and notifications integration for the Dashboard frontend. The UI will inform users when a newer version is available and guide them to upgrade (without performing upgrades).

### Background

The "Upgrade Available" feature is split into three parts: frontend, backend, and interface. This issue covers the **frontend** implementation.

### Deliverables

// TODO

### Requirements

- [ ] Don’t show immediately: wait 3–5 seconds after dashboard is ready
- [ ] If user closes banner: don’t show again for 24 hours
- [ ] If user selects “Don’t remind me”: store in localStorage and don’t show for that specific version
- [ ] Fail silently if version/GitHub data is unavailable (no banner)
- [ ] Cache version checks in localStorage for up to 12 hours (skip requests when cache is fresh; refresh when expired).

DavidReque:
Fixes #927 

## Summary

Add an "Upgrade Available" banner and notifications integration to the Dashboard frontend. When a newer version of the CLI or Helm chart is detected via the backend GraphQL API, users see an informative banner and a notification dot in the footer, guiding them to upgrade using their package manager.

## Changes

- Add `cliVersionStatus` and `clusterVersionStatus` GraphQL queries in `ops.ts` and regenerate types
- Create `upgrade-notifications.ts` module with localStorage caching (12h TTL), dismiss (24h TTL), per-version ignore, and `useUpgradeNotification` hook
- Create `UpgradeBanner` component with informative blue banner, dismiss button, and "Don't remind me" action
- Integrate banner into `AppLayout` with dynamic height adjustment via ResizeObserver
- Update `NotificationsPopover` to show upgrade status with a blue dot indicator and enable it in the Footer
- Add 10 tests covering delay, dismiss, cache, don't-remind-me, fail-silently, and cluster mode behavior

## Submitter checklist

- [X] Add the correct emoji to the PR title
- [X] Link the issue number, if any, to *Fixes #*
- [x] Rebase branch to HEAD

### DIFF
--- a/dashboard-ui/src/lib/update-notifications.tsx
+++ b/dashboard-ui/src/lib/update-notifications.tsx
@@ -64,6 +64,10 @@ function patchState(patch: Partial<UpdateState>) {
   writeState({ ...readState(), ...patch });
 }
 
+function isCacheValid(state: UpdateState): boolean {
+  return !!state.latestVersion && !!state.fetchedAt && Date.now() - state.fetchedAt < CACHE_TTL_MS;
+}
+
 export interface UpdateNotificationState {
   showBanner: boolean;
   currentVersion: string;
@@ -73,16 +77,7 @@ export interface UpdateNotificationState {
   dontRemindMe: () => void;
 }
 
-const defaultState: UpdateNotificationState = {
-  showBanner: false,
-  currentVersion: '',
-  latestVersion: null,
-  updateAvailable: false,
-  dismiss: () => {},
-  dontRemindMe: () => {},
-};
-
-const UpdateNotificationContext = createContext<UpdateNotificationState>(defaultState);
+const UpdateNotificationContext = createContext({} as UpdateNotificationState);
 
 export function UpdateNotificationProvider({ children }: React.PropsWithChildren) {
   const isDesktop = appConfig.environment === 'desktop';
@@ -90,13 +85,7 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
   const [ready, setReady] = useState(false);
   const [state, setState] = useState(() => readState());
 
-  const cachedLatest = useMemo(() => {
-    if (!state.latestVersion || !state.fetchedAt) return null;
-    if (Date.now() - state.fetchedAt >= CACHE_TTL_MS) return null;
-    return state.latestVersion;
-  }, []);
-
-  const skipQuery = !isDesktop || !currentVersion || cachedLatest !== null;
+  const cacheValid = isCacheValid(state);
 
   useEffect(() => {
     const timer = setTimeout(() => setReady(true), SHOW_DELAY_MS);
@@ -104,19 +93,16 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
   }, []);
 
   const { data } = useQuery(CLI_LATEST_VERSION, {
-    skip: skipQuery,
+    skip: !isDesktop || !currentVersion || cacheValid,
     fetchPolicy: 'network-only',
   });
 
-  const latestVersion: string | null = cachedLatest ?? data?.cliLatestVersion ?? null;
-
   useEffect(() => {
-    if (cachedLatest !== null || !isDesktop) return;
     const version = data?.cliLatestVersion;
-    if (version) {
-      patchState({ latestVersion: version, fetchedAt: Date.now() });
-    }
-  }, [cachedLatest, isDesktop, data]);
+    if (version) patchState({ latestVersion: version, fetchedAt: Date.now() });
+  }, [data]);
+
+  const latestVersion = cacheValid ? state.latestVersion! : (data?.cliLatestVersion ?? null);
 
   const updateAvailable =
     currentVersion !== '' && latestVersion !== null && compareSemver(latestVersion, currentVersion) > 0;
@@ -133,12 +119,11 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
   }, []);
 
   const dontRemindMe = useCallback(() => {
-    const current = readState();
-    const versions = current.skippedVersions ?? [];
-    if (latestVersion && !versions.includes(latestVersion)) {
-      versions.push(latestVersion);
+    const { skippedVersions = [] } = readState();
+    if (latestVersion && !skippedVersions.includes(latestVersion)) {
+      skippedVersions.push(latestVersion);
     }
-    writeState({ ...current, skippedVersions: versions, dismissedAt: Date.now() });
+    patchState({ skippedVersions, dismissedAt: Date.now() });
     setState(readState());
   }, [latestVersion]);
 


dashboard-ui/src/components/widgets/Footer.tsx
```diff
@@ -28,13 +28,11 @@ export default function Footer() {
           Settings
         </button>
       </SettingsPopover>
-      {/*
       <NotificationsPopover>
         <button type="button" className="h-full hover:bg-secondary px-1">
           <Bell className="h-4 w-4" />
```
amorey:
```suggestion
          <Bell className="h-3.5 w-3.5" />
```

dashboard-ui/src/components/widgets/NotificationsPopover.tsx
```diff
@@ -12,19 +12,54 @@
...
+
+  const hasCliUpdate = appConfig.environment === 'desktop' && cliStatus?.updateAvailable;
+  const hasClusterUpdate = clusterStatus?.updateAvailable;
+  const hasNotifications = hasCliUpdate || hasClusterUpdate;
 
   return (
     <Popover open={isOpen} onOpenChange={setIsOpen}>
-      <PopoverTrigger asChild>{children}</PopoverTrigger>
+      <PopoverTrigger asChild>
+        <div className="relative">
```
amorey: You need to tell the div to inherit the height of its container otherwise the button will appear top-aligned:
```suggestion
        <div className="relative h-full">
```

dashboard-ui/src/components/widgets/UpdateBanner.test.tsx
amorey, line 1: Let's move this file up a level into the `widgets/` directory to match other tests in the project.

dashboard-ui/src/lib/upgrade-notifications.ts
```diff
@@ -0,0 +1,188 @@
...
+
+export interface UpgradeNotificationState {
+  showBanner: boolean;
+  cliStatus: VersionStatusData | null;
+  clusterStatus: VersionStatusData | null;
+  dismiss: () => void;
+  dontRemindMe: () => void;
+}
+
+export function useUpgradeNotification(kubeContext: string | null): UpgradeNotificationState {
```
amorey: The business logic here (e.g. network fetches, localStorage read+writes) will get executed independently every time `useUpgradeNotifcation()` is used. You could avoid this by using the graphql cache but it would probably be easier to use a Context/Provider architecture instead.

dashboard-ui/src/components/layouts/AppLayout.tsx
```diff
@@ -12,12 +12,35 @@
...
 
 export default function AppLayout({ children }: React.PropsWithChildren) {
+  const [bannerHeight, setBannerHeight] = useState(0);
+
+  const bannerRef = useCallback((node: HTMLDivElement | null) => {
+    if (!node) {
+      setBannerHeight(0);
+      return;
+    }
+    const observer = new ResizeObserver((entries) => {
```
amorey: If you change to a Context/Provider architecture for `useUpgradeNotifications()` then I don't think you need to use a `ResizeObserver` anymore. Instead you can just set up a reactive trigger and render the banner when an upgrade is available.

dashboard-ui/src/components/widgets/NotificationsPopover.tsx
```diff
@@ -12,19 +12,54 @@
...
+import appConfig from '@/app-config';
+import { useUpgradeNotification } from '@/lib/upgrade-notifications';
+
 export const NotificationsPopover = ({ children }: React.PropsWithChildren) => {
   const [isOpen, setIsOpen] = useState(false);
+  const { cliStatus, clusterStatus } = useUpgradeNotification();
+
+  const hasCliUpdate = appConfig.environment === 'desktop' && cliStatus?.updateAvailable;
+  const hasClusterUpdate = clusterStatus?.updateAvailable;
+  const hasNotifications = hasCliUpdate || hasClusterUpdate;
```
chilin0525: Currently, `updateAvailable` is read directly without applying the ignored versions filter used by the upgrade banner. As a result, after clicking "Don't remind me" on the banner, the banner itself disappears correctly, but the bell icon badge in the footer remains visible.

The image below shows the state after the user clicks "Don't remind me" and reloads the page. The notification still appears and continues to display the upgrade notification.

Is this expected behavior?

Image

dashboard-ui/src/components/layouts/AppLayout.tsx
```diff
@@ -13,12 +13,17 @@
...
+import UpgradeBanner from '@/components/widgets/UpgradeBanner';
+import { UpgradeNotificationProvider } from '@/lib/upgrade-notifications';
 
 export default function AppLayout({ children }: React.PropsWithChildren) {
   return (
-    <>
-      <div className="h-[calc(100vh-23px)] overflow-hidden">{children}</div>
-      <Footer />
-    </>
+    <UpgradeNotificationProvider>
```
amorey: It would be more straightforward to lift the `<UpgradeNotificationProvider>` out of `<AppLayout>` into the root (`main.tsx`) and use `useNotificationProvider()` here to render the banner conditionally.

dashboard-ui/src/components/layouts/AppLayout.tsx
```diff
@@ -13,12 +13,17 @@
...
+      <div className="flex h-screen flex-col">
```
amorey: Switching from `calc()` to `flex-1` simplifies the layout but also makes the outer-most container slightly scrollable. Just add `overflow-hidden` to the container to keep it fixed:
```suggestion
      <div className="flex h-screen flex-col overflow-hidden">
```

dashboard-ui/src/components/widgets/NotificationsPopover.tsx
amorey: Personally I think it's ok for the bell icon badge to remain after the banner is dismissed. It's similar to how others handle it (e.g. macOS) and it's subtle anyway.

dashboard-ui/src/components/widgets/UpgradeBanner.tsx
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ArrowUpCircle, X } from 'lucide-react';
+
+import appConfig from '@/app-config';
+import { useUpgradeNotification } from '@/lib/upgrade-notifications';
+
+export default function UpgradeBanner() {
+  const { showBanner, cliStatus, clusterStatus, dismiss, dontRemindMe } = useUpgradeNotification();
```
amorey: Instead of calling `useUpgradeNotification()` here, I'd move the hook call into `<AppLayout>` and conditionally render the banner there. That would make the rendering decision more straightforward and also make this component easier to test.

dashboard-ui/src/lib/upgrade-notifications.tsx
```diff
@@ -0,0 +1,206 @@
...
+interface VersionStatusData {
+  currentVersion: string;
+  latestVersion: string;
+  updateAvailable: boolean;
+}
+
+interface CachedVersionData {
+  timestamp: number;
+  cli: VersionStatusData | null;
+  cluster: VersionStatusData | null;
```
amorey: We need to handle version status for multiple clusters. I think that also means we need to make [`ClusterVersionStatus()`](https://github.com/kubetail-org/kubetail/blob/main/modules/dashboard/graph/schema.resolvers.go#L588-L596) `kubeContext`-aware.

dashboard-ui/src/components/widgets/NotificationsPopover.tsx
DavidReque: Thanks for catching that @chilin0525! I agree with @amorey here. I think it's fine for the bell icon badge to remain visible even after dismissing the banner. It serves as a subtle, non-intrusive reminder that an update is available, similar to how macOS handles update notifications. The banner dismissal and "Don't remind me" are intentionally scoped to the banner itself.

amorey (changes requested):
Your changes look good but I can still see some potential issues with `kubeContext` handling (e.g. `__default__` can change) and also the current behavior will show/hide the banner as the user switches between different clusters which isn't ideal. Also, we haven't added `kubeContext` support to `ClusterVersionStatus()` yet.

To deal with this, here's a proposal:
- Remove cluster upgrade notification handling from this PR and focus exclusively on cli upgrade notifications (using existing banner + bell popover implementation)
- In a separate PR, add `kubeContext` support to `ClusterVersionStatus()` and cluster upgrade notifications to the UI

In terms of the UI for cluster upgrades, maybe we don't need to use the banner and instead we can include it in the health status indicator (also potentially in the bell popover).

Separately, what do you guys think about disabling upgrade notifications for cluster deployments altogether? I think it makes sense to notify desktop clients about cluster version mismatches but cluster deployments are self-consistent so maybe we can let admins handle upgrades on their own without notifying app users in the UI.

chilin0525:
@amorey Sounds good to me.👍

amorey (commented):
Currently, the server fetches both current and latest versions from GitHub on every GraphQL query, does the semver comparison, and returns `{currentVersion, latestVersion, updateAvailable}`. Although this makes the query code simpler, it complicates the caching logic because then you have to bust the cache when there's an upgrade available in order to avoid maintaining a stale `currentVersion` in `localStorage` ([upgrade-notifications.tsx:L55](https://github.com/kubetail-org/kubetail/pull/998/changes#diff-fbd3874461ddace8a4bce662dd14f0b8273fe611b18cd65c6bd2c62b0832ff83R55)).
If I understand the current behavior, when there's an upgrade available every page load will result in a network fetch to GitHub.


Instead, if you split up the code and get `latestVersion` and `currentVersion` separately then you could cache `latestVersion` in `localStorage` and only bust that cache when the ttl expires. In terms of getting the `currentVersion` you could continue getting it server-side or pass it into the client via `appConfig` (and/or do semver comparison client-side).

Longer term, it would be good to cache latest versions server-side so we can use them for CLI calls too. I just looked at how codex does it and they store it in `~.codex/version.json`:
```
{"latest_version":"0.113.0","last_checked_at":"2026-03-10T19:01:12.181601Z","dismissed_version":"0.87.0"}
```

We could do something similar by storing the versions at `~/.kubetail/cache/cli-version.json` and `~/.kubetail/cache/helm-version.json`.

Anyways, I only mention this because if we can architect the current PR in a way that we could replace `localStorage` with server-side storage then it will be easier to switch later.

What do you think?

amorey:
Awesome sounds like a good plan. Let’s use this PR.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -0,0 +1,155 @@
...
+const defaultState: UpdateNotificationState = {
+  showBanner: false,
+  currentVersion: '',
+  latestVersion: null,
+  updateAvailable: false,
+  dismiss: () => {},
+  dontRemindMe: () => {},
+};
+
+const UpdateNotificationContext = createContext<UpdateNotificationState>(defaultState);
```
amorey: **note:** Although `createContext` expects a default value, it isn't actually used anywhere because the value is set at L145 at runtime. What I usually do to avoid creating an unnecessary default state is this hack:

```suggestion
const UpdateNotificationContext = createContext({} as UpdateNotificationState);
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const timer = setTimeout(() => setReady(true), SHOW_DELAY_MS);
+    return () => clearTimeout(timer);
+  }, []);
+
+  const { data } = useQuery(CLI_LATEST_VERSION, {
+    skip: skipQuery,
+    fetchPolicy: 'network-only',
+  });
+
+  const latestVersion: string | null = cachedLatest ?? data?.cliLatestVersion ?? null;
```
amorey: **nitpick:** you should let typescript infer types when possible

```suggestion
  const latestVersion = cachedLatest ?? data?.cliLatestVersion ?? null;
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const [ready, setReady] = useState(false);
+  const [state, setState] = useState(() => readState());
+
+  const cachedLatest = useMemo(() => {
+    if (!state.latestVersion || !state.fetchedAt) return null;
+    if (Date.now() - state.fetchedAt >= CACHE_TTL_MS) return null;
+    return state.latestVersion;
+  }, []);
+
+  const skipQuery = !isDesktop || !currentVersion || cachedLatest !== null;
```
amorey: **suggestion:** It's a bit difficult to follow the cache check/update logic. Here's a suggestion on how to improve it:

```typescript
// Extract the logic into a function
function isCacheValid(state: UpdateState): boolean {
  return !!state.latestVersion && !!state.fetchedAt && Date.now() - state.fetchedAt < CACHE_TTL_MS;
}

// ...inside the provider:
const cacheValid = isCacheValid(state);

const { data } = useQuery(CLI_LATEST_VERSION, {
  skip: !isDesktop || !currentVersion || cacheValid,
  fetchPolicy: 'network-only',
});

useEffect(() => {
  const version = data?.cliLatestVersion;
  if (version) patchState({ latestVersion: version, fetchedAt: Date.now() });
}, [data]);

const latestVersion = cacheValid ? state.latestVersion! : (data?.cliLatestVersion ?? null);
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -0,0 +1,155 @@
...
+    setState(readState());
+  }, []);
+
+  const dontRemindMe = useCallback(() => {
+    const current = readState();
+    const versions = current.skippedVersions ?? [];
+    if (latestVersion && !versions.includes(latestVersion)) {
+      versions.push(latestVersion);
+    }
+    writeState({ ...current, skippedVersions: versions, dismissedAt: Date.now() });
```
amorey: **question:** can we use `patchState()` here?

dashboard-ui/src/pages/home/index.tsx
```diff
@@ -320,13 +321,13 @@ const DisplayWorkloadItems = memo(({ kind }: DisplayWorkloadItemsProps) => {
   const isFetching = useAtomValue(workloadIsFetchingAtomFamilies[kind](kubeContext));
   const items = useAtomValue(filteredWorkloadItemsAtomFamilies[kind](kubeContext));
 
-  const itemIDs = useMemo(() => items.map((item) => item.metadata.uid), [items]);
+  const itemIDs: string[] = useMemo(() => items.map((item: WorkloadItem) => item.metadata.uid), [items]);
```
amorey, line 324: **note:** types already inferred
```suggestion
  const itemIDs = useMemo(() => items.map((item) => item.metadata.uid), [items]);
```

dashboard-ui/src/pages/home/index.tsx
```diff
@@ -320,13 +321,13 @@ const DisplayWorkloadItems = memo(({ kind }: DisplayWorkloadItemsProps) => {
...
 
   const logFileInfo = useLogFileInfo(kubeContext, itemIDs);
 
   const data = useMemo(
     () =>
-      items.map((item) => {
+      items.map((item: WorkloadItem) => {
```
amorey, line 330: **note:** types already inferred
```suggestion
      items.map((item) => {
```

dashboard-ui/src/pages/home/index.tsx
```diff
@@ -77,6 +77,7 @@ import {
   workloadIsFetchingAtomFamilies,
   filteredWorkloadItemsAtomFamilies,
 } from './state';
+import type { WorkloadItem } from './shared';
```
amorey, line 80: **note:** import is uneccessary if we let typescript infer types
```suggestion
```

dashboard-ui/src/lib/update-notifications.test.tsx
```diff
@@ -0,0 +1,166 @@
+// Copyright 2024 The Kubetail Authors
```
amorey, line 1: **praise:** nice tests

amorey:
> I confirmed that the Web Locks API works on localhost (it's a secure context) and you can hold a lock for a tab's lifetime by returning a never-resolving promise. This would work well for leader-election so only one tab fetches from GitHub.
> 
> For cross-tab synchronization of the banner state (dismiss/skip), BroadcastChannel like in AuthProvider is a great fit since the pattern is already established in the codebase.
> 
> I'll track both in a follow-up issue/PR to keep this one focused.

Sounds great

dashboard-ui/src/pages/home/index.tsx
DavidReque: The types aren't actually inferred here — `filteredWorkloadItemsAtomFamilies` returns an untyped atom, so removing the `WorkloadItem` annotation causes Parameter `item` implicitly has an `any` type. I'll leave both annotations and the import as-is. We could fix this upstream by adding proper typing to the atom family, but that's outside the scope of this PR.

amorey: We can deal with this in a separate PR but in my IDE (vscode) I see `filteredWorkloadItemsAtomFamilies` returning `Record<WorkloadKind, AtomFamily<KubeContext, Atom<<number>>>` and `useAtomValue(filteredWorkloadItemsAtomFamilies[kind](kubeContext))` returning `WorkloadItem[]`. If they were untyped I think the build test wouldn't have passed so I'm surprised we're seeing something different. Where are you seeing the error?

Image

Image

DavidReque: You're right — the issue was on my end. `jotai-family` wasn't installed in my local `node_modules`, so TypeScript couldn't resolve the atom family types and everything fell back to any.
I'll apply your suggestions in a follow-up since the PR is already merged. Sorry about that!

amorey: Thats’s interesting… any idea why it didn’t get installed automatically?

DavidReque: It looks like my local frontend deps were stale. This repo doesn’t use a root pnpm workspace, so dashboard-ui dependencies only get installed when I run `pnpm install` inside dashboard-ui. I must have updated the branch without reinstalling there, so TypeScript couldn’t resolve `jotai-family`.

PR #1008, opened 2026-03-19
🐋 Add kubeContext support to ClusterVersionStatus

### LINKED ISSUE #1002: Add cluster upgrade notifications to health status indicator
## Description

Add cluster upgrade notifications to the Dashboard frontend, integrated into the health status indicator (`ServerStatus` widget) and the bell popover, since cluster version status is per-cluster information.

## Background

In the review of #927, the banner approach for cluster notifications was deferred because:

- `__default__` kubeContext can change, making cache keys unreliable
- The banner flickered as users switched between clusters
- `ClusterVersionStatus()` lacked `kubeContext` support on the backend

## Plan

1. A PR to add `kubeContext` support to `ClusterVersionStatus()` on the backend
2. A PR to add cluster upgrade notifications to the frontend UI

## Frontend requirements

- Consume `CLUSTER_VERSION_STATUS` query with the active `kubeContext`
- Show cluster version status in the `ServerStatus` health indicator dialog
- Show cluster upgrade entry in `NotificationsPopover` (bell icon)
- Only activate in desktop mode — cluster deployments should not show upgrade notifications
- Reuse cache/dismiss/ignore logic from `upgrade-notifications.tsx`, keyed per cluster context (12h TTL)
- Fail silently if data is unavailable
- Add tests

## Dependencies

- `ClusterVersionStatus()` must support `kubeContext` on the backend

DavidReque:
Fixes #1002 (backend dependency)

## Summary

Add kubeContext support to `ClusterVersionStatus()` in the dashboard backend so the GraphQL API returns the Helm chart version for the cluster associated with the active kubeContext. This is required for the frontend PR that will show cluster upgrade notifications in the health status indicator and bell popover (per-cluster information in desktop mode).

## Changes

* Extend `helmReleaseGetter` interface with `GetReleaseForContext(kubeContext, namespace, releaseName string)` instead of `GetRelease(namespace, releaseName string)`
* Add `GetReleaseForContext` to `helm.Client`: creates a context-scoped client when `kubeContext` is non-empty, otherwise uses the default context
* Update `ClusterVersionStatus` resolver to resolve the active context via `r.cm.DerefKubeContext(kubeContext)` and pass it to the getter
* Update mock and tests for the new interface and add `ConnectionManager` mock to the Desktop mode tests

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/modules/dashboard/graph/schema.resolvers_test.go
+++ b/modules/dashboard/graph/schema.resolvers_test.go
@@ -34,14 +34,20 @@ import (
 	"github.com/kubetail-org/kubetail/modules/dashboard/pkg/config"
 )
 
-// mockHelmReleaseGetter implements helmReleaseGetter for testing.
-type mockHelmReleaseGetter struct {
-	release *release.Release
-	err     error
+// MockHelmReleaseGetter implements helmReleaseGetter for testing.
+type MockHelmReleaseGetter struct {
+	mock.Mock
 }
 
-func (m *mockHelmReleaseGetter) GetRelease(namespace, releaseName string) (*release.Release, error) {
-	return m.release, m.err
+func (m *MockHelmReleaseGetter) GetReleaseForContext(kubeContext, namespace, releaseName string) (*release.Release, error) {
+	ret := m.Called(kubeContext, namespace, releaseName)
+
+	var r0 *release.Release
+	if ret.Get(0) != nil {
+		r0 = ret.Get(0).(*release.Release)
+	}
+
+	return r0, ret.Error(1)
 }
 
 func makeRelease(name, version string) *release.Release {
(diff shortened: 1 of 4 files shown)


modules/dashboard/graph/helm_release_getter.go
```diff
@@ -0,0 +1,36 @@
...
+type contextAwareHelmReleaseGetter struct {
+	kubeconfigPath string
+}
+
+func (g *contextAwareHelmReleaseGetter) GetReleaseForContext(kubeContext, namespace, releaseName string) (*release.Release, error) {
+	opts := []helm.ClientOption{helm.WithKubeconfigPath(g.kubeconfigPath)}
+	if kubeContext != "" {
+		opts = append(opts, helm.WithKubeContext(kubeContext))
+	}
+	client := helm.NewClient(opts...)
```
amorey: The default `KubeContext` is an empty string anyway so it should be safe to use the option always:
```suggestion
	client := helm.NewClient(helm.WithKubeconfigPath(g.kubeconfigPath), helm.WithKubeContext(kubeContext))
```

modules/dashboard/graph/helm_release_getter.go
amorey, line 1: This should go in `schema.helpers.go` to keep the number of files down and also to place it next to the interface definition.

modules/dashboard/graph/helm_release_getter.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"helm.sh/helm/v3/pkg/release"
+
+	"github.com/kubetail-org/kubetail/modules/shared/helm"
+)
+
+// contextAwareHelmReleaseGetter implements helmReleaseGetter by creating a new
+// helm client per request, scoped to the given kubeContext.
+type contextAwareHelmReleaseGetter struct {
```
amorey: In cases like this when you define an interface just for dependency injection, it's typical to use a name like `defaultHelmReleaseGetter` for the actual struct.

modules/dashboard/graph/schema.resolvers_test.go
amorey: If you use the `testify/mock` library it makes mocking much easier. Here's an example from the `MockConnectionManager`:
https://github.com/kubetail-org/kubetail/blob/a74b901c1df8234bd5fad8269d34d4b1b7caa564/modules/shared/k8shelpers/mock/connection-manager.go#L29-L42

PR #1012, opened 2026-03-26
🐋Add cluster upgrade notifications in server status and bell popover

### LINKED ISSUE #1011: Add cluster upgrade notifications to health status indicator and bell popover
## Description

Add cluster upgrade notifications to the Dashboard frontend, integrated into the health status indicator (ServerStatus widget) and the bell popover, since cluster version status is per-cluster information.

## Background

In the review of #927, the banner approach for cluster notifications was deferred because:

- `__default__` kubeContext can change, making cache keys unreliable
- The banner flickered as users switched between clusters
- `ClusterVersionStatus()` lacked kubeContext support on the backend

The backend dependency has been addressed in [PR #1008](https://github.com/kubetail-org/kubetail/pull/1008).

## Requirements

- Consume `CLUSTER_VERSION_STATUS` query with the active kubeContext
- Show cluster version status in the ServerStatus health indicator dialog
- Show cluster upgrade entry in NotificationsPopover (bell icon)
- Only activate in desktop mode — cluster deployments should not show upgrade notifications
- Reuse cache/dismiss/ignore logic from `update-notifications.tsx`, keyed per cluster context (12h TTL)
- Fail silently if data is unavailable
- Add tests

## Parent 
- Parent: #1002 (Upgrade Notifications)

DavidReque:
Fixes #1011 

## Summary

Adds cluster (Helm chart) upgrade hints in desktop mode: queries `CLUSTER_VERSION_STATUS` per kube context and surfaces it in the health dialog and notifications bell.

## Changes

- `useClusterUpgradeNotification` hook with per-context cache/dismiss (12h TTL, same idea as CLI update notifications).
- Kubetail version row in ServerStatus (desktop only).
- Cluster upgrade line and blue dot on `NotificationsPopover` when a cluster or CLI update is available.
- Tests in `upgrade-notifications.test.tsx`

## Submitter checklist

- [x] Rebase branch to HEAD

### DIFF
--- a/dashboard-ui/src/components/widgets/NotificationsPopover.tsx
+++ b/dashboard-ui/src/components/widgets/NotificationsPopover.tsx
@@ -12,39 +12,70 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { useSubscription } from '@apollo/client/react';
 import { ArrowUpCircle } from 'lucide-react';
 import { useState } from 'react';
 
 import { Popover, PopoverTrigger, PopoverContent } from '@kubetail/ui/elements/popover';
 
-import { useUpdateNotification } from '@/lib/update-notifications';
+import appConfig from '@/app-config';
+import * as dashboardOps from '@/lib/graphql/dashboard/ops';
+import { useCLIUpdateNotification, useClusterUpdateNotification } from '@/lib/update-notifications';
+
+const ClusterUpdateEntry = ({ kubeContext }: { kubeContext: string }) => {
+  const { updateAvailable, currentVersion, latestVersion } = useClusterUpdateNotification(kubeContext);
+
+  if (!updateAvailable || !latestVersion) return null;
+
+  return (
+    <div className="flex items-start gap-2 rounded border border-blue-200 bg-blue-50 p-2 text-sm text-blue-900 dark:border-blue-800 dark:bg-blue-950 dark:text-blue-100">
+      <ArrowUpCircle className="mt-0.5 h-4 w-4 shrink-0" />
+      <p>
+        Cluster update: {currentVersion} → {latestVersion}
+      </p>
+    </div>
+  );
+};
+
+const useActiveKubeContext = (): string | null => {
+  const { data } = useSubscription(dashboardOps.KUBE_CONFIG_WATCH, {
+    skip: appConfig.environment !== 'desktop',
+  });
+  return data?.kubeConfigWatch?.object?.currentContext ?? null;
+};
 
 export const NotificationsPopover = ({ children }: React.PropsWithChildren) => {
   const [isOpen, setIsOpen] = useState(false);
-  const { updateAvailable, currentVersion, latestVersion } = useUpdateNotification();
+  const { updateAvailable, currentVersion, latestVersion } = useCLIUpdateNotification();
+  const activeKubeContext = useActiveKubeContext();
+  const { updateAvailable: clusterUpdateAvailable } = useClusterUpdateNotification(activeKubeContext ?? '');
+
+  const hasCliUpdate = updateAvailable && !!latestVersion;
+  const hasClusterUpdate = appConfig.environment === 'desktop' && !!activeKubeContext && clusterUpdateAvailable;
+  const hasNotifications = hasCliUpdate || hasClusterUpdate;
 
   return (
     <Popover open={isOpen} onOpenChange={setIsOpen}>
       <PopoverTrigger asChild>
         <div className="relative h-full">
           {children}
-          {updateAvailable && <span className="absolute top-0.5 right-0.5 h-2 w-2 rounded-full bg-blue-500" />}
+          {hasNotifications && <span className="absolute top-0.5 right-0.5 h-2 w-2 rounded-full bg-blue-500" />}
         </div>
       </PopoverTrigger>
       {isOpen && (
         <PopoverContent side="top" className="w-80 mr-1">
           <div className="space-y-2">
             <p className="text-sm font-medium">Notifications</p>
-            {updateAvailable && latestVersion ? (
+            {hasCliUpdate && (
               <div className="flex items-start gap-2 rounded border border-blue-200 bg-blue-50 p-2 text-sm text-blue-900 dark:border-blue-800 dark:bg-blue-950 dark:text-blue-100">
                 <ArrowUpCircle className="mt-0.5 h-4 w-4 shrink-0" />
                 <p>
                   CLI update: {currentVersion} → {latestVersion}
                 </p>
               </div>
-            ) : (
-              <p className="text-sm text-muted-foreground">No new notifications</p>
             )}
+            {hasClusterUpdate && <ClusterUpdateEntry kubeContext={activeKubeContext} />}
+            {!hasNotifications && <p className="text-sm text-muted-foreground">No new notifications</p>}
           </div>
         </PopoverContent>
       )}
--- a/dashboard-ui/src/components/widgets/NotificationsPopover.tsx
+++ b/dashboard-ui/src/components/widgets/NotificationsPopover.tsx
@@ -26,9 +26,16 @@ type ClusterUpdateInfo = {
   latestVersion: string | null;
 };
 
+const UpdateNotice = ({ children }: React.PropsWithChildren) => (
+  <div className="flex items-start gap-2 rounded border border-blue-200 bg-blue-50 p-2 text-sm text-blue-900 dark:border-blue-800 dark:bg-blue-950 dark:text-blue-100">
+    <ArrowUpCircle className="mt-0.5 h-4 w-4 shrink-0" />
+    <p>{children}</p>
+  </div>
+);
+
 type ClusterUpdateSubscriberProps = {
   kubeContext: string;
-  onChange: (kubeContext: string, info: ClusterUpdateInfo) => void;
+  onChange: (kubeContext: string, info: ClusterUpdateInfo | null) => void;
 };
 
 /**
@@ -42,7 +49,7 @@ const ClusterUpdateSubscriber = ({ kubeContext, onChange }: ClusterUpdateSubscri
 
   useEffect(() => {
     onChange(kubeContext, { hasUpdate, currentVersion, latestVersion });
-    return () => onChange(kubeContext, { hasUpdate: false, currentVersion: null, latestVersion: null });
+    return () => onChange(kubeContext, null);
   }, [kubeContext, hasUpdate, currentVersion, latestVersion, onChange]);
 
   return null;
@@ -55,8 +62,14 @@ export const NotificationsPopover = ({ children }: React.PropsWithChildren) => {
 
   // Aggregated cluster update state by kubeContext, populated by long-lived subscribers below.
   const [clusterUpdates, setClusterUpdates] = useState<Record<string, ClusterUpdateInfo>>({});
-  const handleClusterUpdateChange = useCallback((kubeContext: string, info: ClusterUpdateInfo) => {
+  const handleClusterUpdateChange = useCallback((kubeContext: string, info: ClusterUpdateInfo | null) => {
     setClusterUpdates((prev) => {
+      if (info === null) {
+        if (!(kubeContext in prev)) return prev;
+        const next = { ...prev };
+        delete next[kubeContext];
+        return next;
+      }
       const cur = prev[kubeContext];
       if (
         cur &&
@@ -94,25 +107,16 @@ export const NotificationsPopover = ({ children }: React.PropsWithChildren) => {
           <div className="space-y-2">
             <p className="text-sm font-medium">Notifications</p>
             {hasCliUpdate && (
-              <div className="flex items-start gap-2 rounded border border-blue-200 bg-blue-50 p-2 text-sm text-blue-900 dark:border-blue-800 dark:bg-blue-950 dark:text-blue-100">
-                <ArrowUpCircle className="mt-0.5 h-4 w-4 shrink-0" />
-                <p>
-                  CLI update: {currentVersion} → {latestVersion}
-                </p>
-              </div>
+              <UpdateNotice>
+                CLI update: {currentVersion} → {latestVersion}
+              </UpdateNotice>
             )}
             {clustersWithUpdates.map((ctx) => {
               const info = clusterUpdates[ctx];
               return (
-                <div
-                  key={ctx}
-                  className="flex items-start gap-2 rounded border border-blue-200 bg-blue-50 p-2 text-sm text-blue-900 dark:border-blue-800 dark:bg-blue-950 dark:text-blue-100"
-                >
-                  <ArrowUpCircle className="mt-0.5 h-4 w-4 shrink-0" />
-                  <p>
-                    Cluster update ({ctx}): {info.currentVersion} → {info.latestVersion}
-                  </p>
-                </div>
+                <UpdateNotice key={ctx}>
+                  Cluster update ({ctx}): {info.currentVersion} → {info.latestVersion}
+                </UpdateNotice>
               );
             })}
             {!hasNotifications && <p className="text-sm text-muted-foreground">No new notifications</p>}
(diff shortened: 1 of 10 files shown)


dashboard-ui/src/components/widgets/NotificationsPopover.tsx
amorey, line 1: Should this code go in the `update-notifications.tsx` file with the previous update code?

amorey (commented):
When I run this branch in development I'm seeing 500 errors from the graphql endpoint and in the UI it says "Checking..." for both live and dead clusters:

Image

Also, instead of adding a new row to the health check UI I think it would be better to set the dot color to blue and set the status string to "Update available".

Additionally, I think we should get rid of the Install button. It's buggy to begin with and users might not even have the permissions to install things in the cluster so it might not work. For now I think it's ok just to let users know that an update is available and let them manage the upgrade themselves.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -138,3 +138,132 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
 export function useUpdateNotification(): UpdateNotificationState {
   return useContext(UpdateNotificationContext);
 }
+
+const CLUSTER_STORAGE_PREFIX = 'kubetail:updates:cluster:';
+
+interface ClusterUpgradeState {
```
amorey: You can use inheritance here:
```typescript
interface UpdateState {
  latestVersion?: string;
  fetchedAt?: number;
  dismissedAt?: number;
  skippedVersions?: string[];
}

interface ClusterUpdateState extends UpdateState {
  currentVersion?: string;
}
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -138,3 +138,132 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
...
+
+function patchClusterState(kubeContext: string, patch: Partial<ClusterUpgradeState>) {
+  writeClusterState(kubeContext, { ...readClusterState(kubeContext), ...patch });
+}
+
+function isClusterCacheValid(state: ClusterUpgradeState): boolean {
+  return state.fetchedAt !== undefined && Date.now() - state.fetchedAt < CACHE_TTL_MS;
+}
+
+export interface ClusterUpgradeNotificationState {
```
amorey: You can use inheritance here to to consolidate `UpdateNotificationState` and `ClusterUpgradeNotificationState`

amorey: We should rename "Upgrade" -> "Update" everywhere to be consistent

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -138,3 +138,132 @@ export function UpdateNotificationProvider({ children }: React.PropsWithChildren
...
+  upgradeAvailable: boolean;
+  currentVersion: string | null;
+  latestVersion: string | null;
+  dismiss: () => void;
+  dontRemindMe: () => void;
+}
+
+export function useClusterUpgradeNotification(kubeContext: string): ClusterUpgradeNotificationState {
```
amorey: Every use of `useClusterUpgradeNotification()` will execute this code which creates race conditions and potentially multiple simultanous requests. Instead we can use `UpdateNotificationContext` to centralize requests behind `UpdateNotificationProvider` like we did with `useUpdateNotification`.

amorey (changes requested):
Sorry for the delay. I added some comments inline. In addition, it looks like there's a lot of duplicated code in `update-notifications.tsx`. I wonder if you can simplify the code by using inheritance and sharing `UpdateNotificationProvider` and the `useUpdateNotification()` hook for both CLI and cluster updates.

dashboard-ui/src/pages/home/index.test.tsx
amorey, line 1: Why do we need to add `<UpdateNotificationProvider>` to the home page test?

amorey (changes requested):
There are some merge conflicts with `main` that would probably be easier for me to handle. Let's address the home page issue I mentioned then collapse this into one commit. Then I can do a rebase on top of that.

dashboard-ui/src/pages/home/index.test.tsx
DavidReque: We don't need `<UpdateNotificationProvider>` in the home test anymore. `useUpdateNotification`(kubeContext) now returns an inactive default state when used outside the provider instead of throwing, so unrelated tests work without extra setup.

Also rebased on main and collapsed everything into a single commit. The merge conflict in `update-notifications.tsx` was just the now state variable you added — incorporated it into our refactored code.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+      updateAvailable: cliUpdateAvailable,
+      dismiss: dismissCli,
+      dontRemindMe: dontRemindCli,
+    }),
+    [showBanner, currentVersionCli, latestVersionCli, cliUpdateAvailable, dismissCli, dontRemindCli],
   );
 
-  return <UpdateNotificationContext.Provider value={value}>{children}</UpdateNotificationContext.Provider>;
+  return (
+    <UpdateNotificationContext.Provider value={cliValue}>
```
amorey: We should make it clear that this context provider is only for CLI context:
```suggestion
    <CLIUpdateNotificationContext.Provider value={cliValue}>
```

dashboard-ui/src/lib/update-notifications.tsx
amorey, line 1: I think using "CLI" instead of "Cli" would be more idomatic

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+}
 
 export function UpdateNotificationProvider({ children }: React.PropsWithChildren) {
   const isDesktop = appConfig.environment === 'desktop';
-  const currentVersion = appConfig.cliVersion;
+  const currentVersionCli = appConfig.cliVersion;
   const [ready, setReady] = useState(false);
-  const [state, setState] = useState(() => readState());
+  const [cliPersisted, setCliPersisted] = useState(() => readCliState());
   const [now] = useState(() => Date.now());
```
amorey: In retrospect it would have been better to name this variable `mountedAt`:
```suggestion
  const [mountedAt] = useState(() => Date.now());
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+function isCliQueryCacheValid(state: UpdateState): boolean {
   return !!state.latestVersion && !!state.fetchedAt && Date.now() - state.fetchedAt < CACHE_TTL_MS;
 }
 
-export interface UpdateNotificationState {
-  showBanner: boolean;
-  currentVersion: string;
+function isClusterFetchedCacheValid(state: ClusterUpdateState): boolean {
+  return state.fetchedAt !== undefined && Date.now() - state.fetchedAt < CACHE_TTL_MS;
+}
```
amorey: If these functions serve the same purpose and signature they should have similar names (e.g. `isCLICacheValid()`, `isClusterCacheValid()`)

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
-  const cacheValid = isCacheValid(state);
+  const cliCacheValid = isCliQueryCacheValid(cliPersisted);
+
+  const { data: kubeData } = useSubscription(KUBE_CONFIG_WATCH, {
+    skip: appConfig.environment !== 'desktop',
+  });
+
+  const kubeContexts = useMemo(() => {
+    if (appConfig.environment === 'cluster') return [''];
+    if (appConfig.environment !== 'desktop') return [];
```
amorey: Currently the only values for environment are `cluster` and `desktop` so this line shouldn't be reached. I think we should remove this to make the code more readable.

Separately, we should consider turning `appConfig.environment` into an enum in a separate PR.

```suggestion
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+  const currentVersionCli = appConfig.cliVersion;
   const [ready, setReady] = useState(false);
-  const [state, setState] = useState(() => readState());
+  const [cliPersisted, setCliPersisted] = useState(() => readCliState());
   const [now] = useState(() => Date.now());
 
-  const cacheValid = isCacheValid(state);
+  const cliCacheValid = isCliQueryCacheValid(cliPersisted);
+
+  const { data: kubeData } = useSubscription(KUBE_CONFIG_WATCH, {
```
amorey: Currently, we have Apollo client configured with [`queryDeduplication: false`](https://github.com/kubetail-org/kubetail/blob/main/dashboard-ui/src/apollo-client.ts#L203) so every `KUBE_CONFIG_WATCH` creates a new subscription on the server (e.g. `<ServerStatus>`, `<KubeContextPicker>`). In a follow-up PR we should consolidate the queries into a Provider/Context/Hook (or consider turning on `queryDeduplication`) to avoid duplication.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+  kubeContext,
+  bumpCluster,
+  setSnapshot,
+}: {
+  kubeContext: string;
+  bumpCluster: () => void;
+  setSnapshot: (ctx: string, snap: ClusterVersionQuerySnapshot) => void;
+}) {
+  const isDesktop = appConfig.environment === 'desktop';
+  const persisted = readClusterState(kubeContext);
```
amorey: Use a state to avoid reading from localStorage on every render:
```suggestion
  const [persisted] = useState(() => readClusterState(kubeContext));
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+      latestVersion?: string | null;
+      updateAvailable: boolean;
+    } | null;
+  } | null;
+  error?: Error;
+  loading: boolean;
+};
+
+type ClusterNotificationsRegistry = {
+  clusterRevision: number;
```
amorey: "cluster" is redundant:
```suggestion
  revision: number;
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+  querySnapshots: Record<string, ClusterVersionQuerySnapshot>;
+  bumpCluster: () => void;
```
amorey: Is it necessary for the `bumpCluster()` method to be part of the registry?

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+};
+
 const UpdateNotificationContext = createContext({} as UpdateNotificationState);
```
amorey:
```suggestion
const CLIUpdateNotificationContext = createContext({} as UpdateNotificationState);
```

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+        return prev;
+      }
+      return { ...prev, [kubeContext]: snap };
+    });
+  }, []);
+
+  const clusterRegistry = useMemo(
+    () => ({ clusterRevision, querySnapshots, bumpCluster }),
+    [clusterRevision, querySnapshots, bumpCluster],
+  );
```
amorey: It looks like you're using `clusterRevision` to trigger context updates but this shouldn't be necessary because you can use `querySnapshots` updates to do the same.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
 }
 
-export function useUpdateNotification(): UpdateNotificationState {
-  return useContext(UpdateNotificationContext);
+export function useUpdateNotification(): UpdateNotificationState;
+export function useUpdateNotification(kubeContext: string): ClusterUpdateNotificationState;
+export function useUpdateNotification(kubeContext?: string): UpdateNotificationState | ClusterUpdateNotificationState {
+  const cli = useContext(UpdateNotificationContext);
+  const clusterRegistry = useContext(ClusterNotificationsContext);
```
amorey: In this hook you're adding a dependency on both contexts which means an update in one of them will trigger a re-render of the hook even in subscribers that only care about the other. Neither context will update very frequently so it's not necessarily bad but it negates the benefit of keeping the contexts separate to begin with.

dashboard-ui/src/lib/update-notifications.tsx
```diff
@@ -31,111 +32,359 @@ interface UpdateState {
...
+        {children}
+      </ClusterNotificationsContext.Provider>
+    </UpdateNotificationContext.Provider>
+  );
 }
 
-export function useUpdateNotification(): UpdateNotificationState {
-  return useContext(UpdateNotificationContext);
+export function useUpdateNotification(): UpdateNotificationState;
+export function useUpdateNotification(kubeContext: string): ClusterUpdateNotificationState;
```
amorey: Currently, any updates to the cluster registry will trigger a re-render of all `useUpdateNotication()` subscribers so even if a caller only subscribes to one `kubeContext` they are still effectively listening for changes in all the others.

To deal with this I think we have two options:
1) Create a hook that returns all cluster updates (e.g. `useClusterUpdateNotications()`)
2) Use Jotai to make targeted listeners

I'm not necessarily in favor of one over the other. Jotai will add complexity in some ways but also make other parts of the code simpler (e.g. registry updates). Keeeping things as-is will be quicker to implement but this is basically the point where you start running into the limits of React Context.

What do you think?

amorey (changes requested):
I added some comments inline but I think the bigger thing to focus on is the context/hook issue (Jotai vs. React Context). We should make a decision about that first because it will affect the overall architecture and may make my other suggestions moot.

dashboard-ui/src/lib/update-notifications.tsx
amorey, line 1: Also, more comments would be helpful. It's hard to see all the different functions/components and also follow some of the longer render methods.

dashboard-ui/src/lib/update-notifications.tsx
DavidReque: I agree this is where React Context starts to show its limits.

I split the CLI and cluster hooks so CLI-only consumers no longer subscribe to cluster updates, but you're right that cluster consumers are still all keyed off the same registry and will re-render when any kubeContext changes.

Between the two options, I think `useClusterUpdateNotifications()` would be the smaller follow-up and keeps the implementation simpler, but it still leaves the caller responsible for selecting from a shared snapshot. If we want truly targeted subscriptions, Jotai is probably the better long-term fit since it maps more naturally to per-kubeContext state and would also simplify the registry/update flow.

For this PR I'd prefer to keep the current approach minimal and avoid a larger architectural shift mid-review, then follow up with a focused PR to decide between:
1. a `useClusterUpdateNotifications()` API as the lighter option, or
2. moving cluster update state to Jotai if we want per-context subscriptions and cleaner state updates.

DavidReque:
> I added some comments inline but I think the bigger thing to focus on is the context/hook issue (Jotai vs. React Context). We should make a decision about that first because it will affect the overall architecture and may make my other suggestions moot.

I think the current PR is in a better place now, but I agree the broader architecture question is still open.

What I addressed here was the immediate React Context issues:
- split CLI and cluster into separate hooks/contexts,
- removed some unnecessary invalidation state,
- and simplified the current implementation.

That said, the deeper concern you raised still stands: cluster consumers are still backed by a shared registry, so this does not fully solve the targeted subscription problem. In that sense, I’d say this PR improves the current React Context approach, but it does not fully settle the Context vs. Jotai decision.

My preference for this PR would be to keep the current implementation minimal and shippable, then do a focused follow-up to decide whether we want:
1. to stay with React Context and reshape the API around shared cluster state, or
2. move the cluster side to Jotai for more targeted subscriptions.

What do you think?

amorey:
@codex review

PR #1093, opened 2026-04-25
🐋 Add cluster-api log download

amorey:
## Summary

The download feature added in #1090 only worked when the dashboard
fetched logs through the Kubernetes API. When the frontend was configured
to use the cluster-api for log fetching (the `shouldUseClusterAPI` path),
downloads still hit the dashboard endpoint and missed the cluster-api's
gRPC log fetcher and `allowedNamespaces` enforcement. This PR aligns
both backends on `POST /api/v1/download`, adds the matching endpoint to
the cluster-api server, and routes the frontend through the cluster-api
proxy when that mode is active.

## Key Changes

- Extract the format-agnostic download core (form binding, validation,
  TSV/CSV/text writers, ANSI stripping, stream-options builder) to
  `modules/shared/logs` so the dashboard and cluster-api handlers share
  it instead of duplicating ~450 lines.
- Rename the dashboard route from `/api/logs/download` to
  `/api/v1/download`. Add the same route on the cluster-api server,
  wired with the gRPC agent log fetcher and configured allowed
  namespaces.
- Frontend `download.tsx` now resolves the form action via a new
  `getDownloadActionURL` helper. When `shouldUseClusterAPI` is true the
  request goes through `cluster-api-proxy/api/v1/download` (in cluster
  mode) or `cluster-api-proxy/<ctx>/<ns>/<svc>/api/v1/download` (in
  desktop mode); otherwise it goes to the dashboard endpoint.
- New `clusterAPIProxyPath` helper plus `CLUSTER_API_PROXY_NAMESPACE` /
  `CLUSTER_API_PROXY_SERVICE` constants in `lib/util.ts` unify the
  proxy basepath logic that was previously duplicated across
  `apollo-client.ts`, `download.tsx`, `console/index.tsx`, and
  `log-metadata-provider.tsx`.
- Bump the cluster-api server `WriteTimeout` from 10s to 1m so streamed
  downloads aren't truncated, mirroring the dashboard bump from #1090.
- Apply `requireTokenMiddleware` to the cluster-api `/api/v1/download`
  route only. `/graphql` keeps the per-resolver token check so graphiql
  introspection remains unauthenticated.
- Harden the bearer-token presence check in both backends'
  `authenticationMiddleware`: trim at the trust boundary so a
  whitespace-only `Authorization: Bearer    ` header is treated as
  absent. Kubernetes always rejected such tokens at the API call, but
  the dashboard's `AuthModeToken` gate was failing open and letting
  requests reach handler code first.
- Cluster-api download handler ignores any `kubeContext` carried in the
  form, since `InClusterConnectionManager` rejects non-empty
  `kubeContext` values outright.

## Checklist

- [x] Related issue linked above, if any (none)

### DIFF
--- a/dashboard-ui/src/apollo-client.ts
+++ b/dashboard-ui/src/apollo-client.ts
@@ -24,7 +24,7 @@ import toast from 'react-hot-toast';
 import appConfig from '@/app-config';
 import clusterAPI from '@/lib/graphql/cluster-api/__generated__/introspection-result.json';
 import dashboard from '@/lib/graphql/dashboard/__generated__/introspection-result.json';
-import { getBasename, joinPaths } from '@/lib/util';
+import { clusterAPIProxyPath, getBasename, joinPaths } from '@/lib/util';
 
 /**
  * Shared items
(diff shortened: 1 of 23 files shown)


amorey:
@codex review

PR #1095, opened 2026-04-25
🎣 Fix log viewer column widths under-reporting and compounding

amorey:
## Summary

Log viewer column widths were measured directly off the cell `<div>`. That broke in two ways: the cell's width is constrained by its grid track (so it under-reported when the track clamped it) and by its own `minWidth` from a previous measurement (so it could compound across renders, with the previous `minWidth` becoming the new measurement). The result was columns that grew unbounded or rendered too narrow depending on layout state.

## Key Changes

- Wrap each cell's content in an inner `<span class="inline-block">` and measure that span instead of the cell. The inline-block span sizes to its intrinsic content, independent of the surrounding grid track or the cell's `minWidth`.
- Add the cell's horizontal padding (`px-2` → 16px) back to the measured content width to derive the cell's required width.
- Skip measurement entirely for the Message column when wrap mode is on, since that column intentionally fills available space.

## Checklist


amorey:
@codex review

PR #1096, opened 2026-04-26
✨ Harden cluster-api authentication and transport

amorey:
## Summary

The cluster-api is intended for bot/programmatic clients, not browsers, so the previous `Sec-Fetch-Site` WebSocket check was the wrong primitive. This PR replaces that with strict per-resolver token enforcement and a CSWSH defense that leans on the dashboard's reverse proxy. It also adds an SSE transport so authenticated subscriptions work from a browser (which can't set headers on a WebSocket upgrade).

## Key Changes

- WebSocket `CheckOrigin` now accepts requests with no `Origin` (bots) and rejects when one is set (direct browser connections). The dashboard's proxy strips `Origin` after `csrfProtectionMiddleware` enforces same-origin, so legitimate dashboard-fronted traffic still passes.
- `LogMetadataList` and `LogMetadataWatch` now call `getBearerTokenRequired`, matching `LogRecordsFetch` and `LogRecordsFollow`.
- Remove `cfg.CSRF`, its yaml section, the README row, and the now-unused `cfg` arg to `graph.NewServer`.
- Register an SSE GraphQL transport so browsers can run subscriptions over authenticated POSTs; treat SSE streams as long-lived for graceful shutdown and drop the server `WriteTimeout` that would otherwise cut streams short.
- Update the GraphiQL playground to route subscriptions through `graphql-sse` while keeping queries and mutations on HTTP.
- Tests: app-level integration test confirms `/graphql` rejects an unauthenticated `logRecordsFetch` while introspection still passes; cluster-api tests cover `CheckOrigin` accept/reject and SSE transport delivery; dashboard proxy tests confirm `Origin` is stripped on both `InClusterProxy` and `DesktopProxy` paths.

## Checklist

### DIFF
--- a/config/default/cluster-api.yaml
+++ b/config/default/cluster-api.yaml
@@ -100,20 +100,6 @@ cluster-agent:
     #
     server-name: ""
 
-## csrf ##
-#
-# CSRF protection settings
-#
-csrf:
-
-  ## enabled ##
-  #
-  # Whether CSRF protection is enabled.
-  #
-  # Default value: true
-  #
-  enabled: true
-
 ## logging ##
 #
 # Configuration for the API server's logging output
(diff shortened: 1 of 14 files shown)


amorey:
@codex review

amorey:
@codex review

amorey:
@codex review

PR #1121, opened 2026-04-29
🎣 Fix context cancellation in mergeLogStreams

jerome-wilson:
Closes #1048 

Add context cancellation checks to both blocking channel reads in mergeLogStreams to ensure the function properly responds to context cancellation at all blocking points. This fixes an intermittent test failure where the goroutine could block indefinitely on a channel read even after context cancellation.

## Summary

Fix intermittent test failure in `TestMergeLogStreamsContextCancellation` caused by missing context cancellation checks in the `mergeLogStreams` function. When the goroutine was blocked on a channel read, it could not respond to context cancellation, causing the test to fail intermittently and potentially causing goroutine leaks in production.

## Key Changes

- Add `select` with `ctx.Done()` check to initialization loop channel read
- Add `select` with `ctx.Done()` check to main loop channel read  
- Close input channel in existing test for proper cleanup
- Add new test `TestMergeLogStreamsContextCancellationWhileBlocking` to verify cancellation works while blocking on input

## Checklist

### DIFF
--- a/modules/shared/logs/helpers.go
+++ b/modules/shared/logs/helpers.go
@@ -237,18 +237,16 @@ func mergeLogStreams(ctx context.Context, reverse bool, streams ...<-chan LogRec
 		// Initialize the heap with the first entry from each stream
 		for _, ch := range streams {
 			// Read one entry if available, with context cancellation check
-			var entry LogRecord
-			var ok bool
 			select {
 			case <-ctx.Done():
 				return
-			case entry, ok = <-ch:
-			}
-			if ok {
-				heap.Push(pq, recordWithSource{
-					record: entry,
-					srcCh:  ch,
-				})
+			case entry, ok := <-ch:
+				if ok {
+					heap.Push(pq, recordWithSource{
+						record: entry,
+						srcCh:  ch,
+					})
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


modules/shared/logs/helpers.go
```diff
@@ -236,8 +236,14 @@ func mergeLogStreams(ctx context.Context, reverse bool, streams ...<-chan LogRec
 
 		// Initialize the heap with the first entry from each stream
 		for _, ch := range streams {
-			// Read one entry if available
-			entry, ok := <-ch
+			// Read one entry if available, with context cancellation check
```
amorey, line 239: I think it reads cleaner if you put the heap push in the `select {}` call instead of registering local vars outside it:
```go
select {
case <-ctx.Done():
  return
case entry, ok = <-ch:
  if ok {
    headp.Push(pq, recordWithSource{
      record: entry,
      srcCh:  ch,
    })
  }
```

Also, the code is clear enough that I don't think you need a comment.

modules/shared/logs/helpers.go
```diff
@@ -258,8 +264,14 @@ func mergeLogStreams(ctx context.Context, reverse bool, streams ...<-chan LogRec
 			case outCh <- earliest.record:
 			}
 
-			// Read the next entry from the same source channel
-			entry, ok := <-earliest.srcCh
+			// Read the next entry from the same source channel, with context cancellation check
```
amorey, line 265: Same comment as above about putting the `heap.Push()` call in the `select {}` statement.

modules/shared/logs/helpers_test.go
amorey: Can we remove the timeout?

amorey (changes requested):
Looking good! I added some suggestions in-line.

amorey (approved):

PR #1135, opened 2026-05-05
🎣 Fix kubetail logs backend fallback, filters, and pagination

amorey:
## Summary

The `kubetail logs` command uses a Kubetail API backend when available, but several behaviors diverged from what users would expect. This PR fixes three classes of issues and adds support for source filters on the kubetail backend.

## Key Changes

- **`--backend=auto` now decides via the actual logs request, not a separate probe.** Previously the CLI checked the kube-aggregator's `APIService` object before issuing a request — which required `get apiservices.apiregistration.k8s.io` for least-privilege users. The CLI now attempts the logs request; an HTTP 404 from the kube-apiserver maps to a typed `ErrAPINotInstalled` and triggers a silent fallback (with a yellow warning) to the kubernetes API backend. Other errors are surfaced. `--backend=kubetail` still reports a friendly error when the API isn't installed.

- **Stream mode is resolved once.** `logsCmdConfig.resolvedMode` is now the single source of truth for `--head`/`--tail`/`--all`/`--since` precedence. Previously, `--since PT30M` (with no explicit mode flag) was interpreted as head mode for the kubernetes backend but fell through to tail on the kubetail backend.

- **HEAD pagination respects `--head=N`.** `Stream.run` paginates only when an explicit `Paginate` flag is set (only `--all`). Previously `--head=N --backend=kubetail` walked every page in N-record chunks instead of stopping after the first page.

- **Page boundaries no longer skip a record.** `PaginateLogRecords` (in `modules/shared/logs`) now uses the boundary of the *returned* slice as `nextCursor` instead of the dropped extra record. Combined with the resolver's exclusive `after`/`before` filter, this prevents losing one record at every page boundary on `--all`. The dashboard is unaffected because it doesn't iterate cursors.

- **Source filters work on the kubetail backend.** `--region`, `--zone`, `--os`, `--arch`, `--node` now flow through to the cluster-api via the `LogSourceFilter` GraphQL input. Previously these flags were silently ignored on the kubetail backend.

## Checklist


amorey:
@codex review

amorey:
@codex review

amorey:
@codex review

amorey:
@codex review

PR #1137, opened 2026-05-05
✨ Add background update check with notification

amorey:
## Summary

Adds an opt-out background update check to the CLI so users are notified when a newer release is available. The check polls the GitHub releases API in the background while the user's command runs and prints a one-line notification on exit if a newer version is found.

## Key Changes

- New `modules/cli/internal/updatecheck` package implementing the GitHub release lookup, semver comparison, on-disk caching, and async refresh, with full unit test coverage.
- Wired into the root command via `PersistentPreRunE`/`PersistentPostRunE` so the check runs once per invocation without blocking command execution.
- Added a `general.update_check` config option (enabled by default) to allow disabling the behavior.

## Checklist

### DIFF
--- a/modules/cli/cmd/root.go
+++ b/modules/cli/cmd/root.go
@@ -15,15 +15,20 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


amorey:
@codex review

PR #1138, opened 2026-05-07
🎣 Fix race condition in stream forwarder causing flaky test

### LINKED ISSUE #1136: Intermittent test failure when testing tail streaming
```
<nil>
&{{0 0 <nil>} {0 0 <nil>} false  <nil> 0xc00044d540 0x554060 0 0 0xc00027cf60 16384 test 0xc0002d6f20 0xc0004d4980 false {{} {{} {} 0} 0} 0xc0002eee70 0xc0002eeee0 0xc0002eef50 <nil> {{} {0 0}} {{} {{} 0} {{} {0 0}}} {{} {{} 0} {{} {0 0}}} {{} {{} 0} {{} {0 0}}} {{} {{} 0} {{} {0 0}}}}
--- FAIL: TestStreamTailWithFollow (0.02s)
    --- FAIL: TestStreamTailWithFollow/with_follow_data,_past_data_arrives_after (0.00s)
        stream_test.go:948: 
            	Error Trace:	/home/runner/work/kubetail/kubetail/modules/shared/logs/stream_test.go:948
            	Error:      	Not equal: 
            	            	expected: []string{"s1-a", "s2-a", "s1-b", "s2-b"}
            	            	actual  : []string{"s1-a", "s2-a", "s2-b", "s1-b"}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -3,4 +3,4 @@
            	            	  (string) (len=4) "s2-a",
            	            	- (string) (len=4) "s1-b",
            	            	- (string) (len=4) "s2-b"
            	            	+ (string) (len=4) "s2-b",
            	            	+ (string) (len=4) "s1-b"
            	            	 }
            	Test:       	TestStreamTailWithFollow/with_follow_data,_past_data_arrives_after
```

jerome-wilson:
Closes #1136 

Records from multiple sources could arrive in non-deterministic order when multiple goroutines write to futureCh simultaneously. This adds a micro-batching mechanism that collects records for a short window and sorts them by timestamp before outputting.
This is similar to the issue fixed in #1121 but affects a different code path in `runForwarder()`.
Fixes flaky TestStreamTailWithFollow test.

## Summary

Fix a race condition in `startFollow_UNSAFE()` where multiple goroutines write to a shared `futureCh` channel independently. When records arrive simultaneously from different sources, the order depends on which goroutine wins the CPU race, causing the test `TestStreamTailWithFollow/with_follow_data,_past_data_arrives_after` to fail intermittently.

## Key Changes

- Added micro-batching in Step 4 of `runForwarder()` to collect records for a short window (5ms)
- Sort buffered records by timestamp before outputting
- Ensures records from different sources that arrive within the same time window are output in correct timestamp order

## Checklist

### DIFF
--- a/modules/shared/logs/stream_test.go
+++ b/modules/shared/logs/stream_test.go
@@ -515,7 +515,8 @@ func TestStreamAllWithFollow(t *testing.T) {
 		setPastStreamAfter1  []LogRecord
 		setPastStreamAfter2  []LogRecord
 		setFutureStream      []LogRecord
-		wantLines            []string
+		wantPastLines        []string
+		wantFutureLines      []string
 	}{
 		{
 			"no follow data, past data arrives before",
@@ -525,6 +526,7 @@ func TestStreamAllWithFollow(t *testing.T) {
 			[]LogRecord{},
 			[]LogRecord{},
 			[]string{"s1-a", "s2-a"},
+			[]string{},
 		},
 		{
 			"no follow data, past data arrives after",
@@ -534,6 +536,7 @@ func TestStreamAllWithFollow(t *testing.T) {
 			[]LogRecord{logs2[0]},
 			[]LogRecord{},
 			[]string{"s1-a", "s2-a"},
+			[]string{},
 		},
 		{
 			"no follow data, past data arrives before and after",
@@ -543,6 +546,7 @@ func TestStreamAllWithFollow(t *testing.T) {
 			[]LogRecord{logs2[1]},
 			[]LogRecord{},
 			[]string{"s1-a", "s2-a", "s1-b", "s2-b"},
+			[]string{},
 		},
 		{
 			"with follow data, past data arrives before",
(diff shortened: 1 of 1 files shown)


amorey (changes requested):
I don't think we should use microbatching to fix this because the timestamps the code has access to are server-reported timestamps anyway so between clock jitter and network latency we can't guarantee real ordering anyway. Instead I think the issue is just a race condition in the tests here:
https://github.com/kubetail-org/kubetail/blob/ec720190aa4d8d1e4b30f08f2602a96f145f5952/modules/shared/logs/stream_test.go#L918-L926

The two channels are competing to write data to the stream and when `ch2New` wins the data isn't in the order that we expect based on `wantLines`.

Maybe the way to solve the problem is to replace `wantLines` with a more deterministic check that asserts the order of past data using `setPastStream*` and then a more relaxed check on future data based on the content but not the ordering of `setFutureStream`.

Incidentally I think the previous test has the same issue:
https://github.com/kubetail-org/kubetail/blob/ec720190aa4d8d1e4b30f08f2602a96f145f5952/modules/shared/logs/stream_test.go#L682-L690

jerome-wilson:
Yes. That makes sense. Instead of implementing micro-batching in production code, I can update the tests to use strict assertions for past data and relaxed assertions for future data.

amorey (approved):
Very nice solution

PR #1139, opened 2026-05-07
🎣 Fix thin-border selection bug and refactor code

amorey:
## Summary

Fixes a long-standing visual bug where cell-selection borders rendered
as thin (1px-looking) lines at the edges of non-rectangular selections —
especially visible on the "step" boundaries of L-shaped selections built
via Cmd+click.

The root cause was per-cell inset box-shadows: stacked inset shadows
have a sub-pixel rendering quirk (most prominent in Firefox), and
adjacent selected rows raced for shared boundary pixels via z-index,
with each "fix" exposing a new mismatch on the other side. Replacing
inset shadows with a real overlay layer eliminates the conflict
entirely — the selection chrome paints as one layer above all cells,
with no row-vs-row painting race.

Along the way, a few related cleanups in the selection subsystem.

## Key Changes

- **Overlay-based selection borders.** New `SelectionOverlay` component
  lives inside each `RecordRow` and draws per-edge real CSS borders on
  absolutely-positioned empty divs. The anchor cell renders as the same
  rect with all four borders forced and a corner-dot indicator. ColorDot
  becomes a clean interior connector when both selectable neighbors are
  selected. `cell-styles.ts` and its tests are deleted.
- **Selection state migrated to Jotai's `useStore()`.** The five
  `useRef` mirrors plus the syncing `useEffect` in `useSelectionState`
  are gone; handlers read via `store.get(atom)` and write via
  `store.set(atom, v)`. This also eliminates a foot-gun in the
  arrow-key handler that required manually pre-writing refs to defeat
  React's batching lag.
- **Shared drag-loop helper.** `useRowDrag` and `useCellDrag` each had
  ~25 lines of identical `requestAnimationFrame` coalescing,
  `AbortController` lifecycle, and `mousemove`/`mouseup` wiring.
  Lifted into `startDocumentDrag<T>`; each caller plugs in its own
  `resolveTarget` / `onMove` / `onCommit`.
- **Spreadsheet-anchor invariant.** Renamed `lastClickedCell` to
  `anchorCell` and tightened the rule: the anchor always points at a
  selected cell (or null). Cmd+clicking the anchor now moves it to the
  next selected cell in reading order. Shift+click replaces the
  selection with a rectangle from the anchor (was: merged with prior
  selection).
- **Per-row drag perf.** `computeCellRange` now shares a single col
  `Set` across all rows in a drag rectangle so unchanged rows keep
  reference-stable selection sets and skip re-render via `RecordRow`'s
  memo.
- **Width-measurement cascade fix.** `useMeasureWidths` only allocates
  a new `colWidths` Map when widths actually changed (was: every flush
  invalidated every `RecordRow` memo).
- **Smaller cleanups.** Extract `clearRowSelection` /
  `clearCellSelection` / `selectSingleCell` helpers; bind the
  cursor-text one-shot listener to an `AbortController` cleaned up on
  unmount; `ARROW_KEYS` constant + `isArrowKey` type guard; export and
  reuse `isSelectableViewerColumn`; drop redundant JSDoc on copy
  formatters.

## Checklist

### DIFF
--- a/dashboard-ui/src/pages/console/main.test.tsx
+++ b/dashboard-ui/src/pages/console/main.test.tsx
@@ -201,6 +201,7 @@ describe('Main', () => {
       selectedCellCols: undefined as Set<ViewerColumn> | undefined,
       selectedCellColsAbove: undefined as Set<ViewerColumn> | undefined,
       selectedCellColsBelow: undefined as Set<ViewerColumn> | undefined,
+      anchorCol: undefined as ViewerColumn | undefined,
       isCursorText: true,
       isCellTextSelectable: false,
       measureElement: vi.fn(),
@@ -261,6 +262,7 @@ describe('Main', () => {
       selectedCellCols: undefined as Set<ViewerColumn> | undefined,
       selectedCellColsAbove: undefined as Set<ViewerColumn> | undefined,
       selectedCellColsBelow: undefined as Set<ViewerColumn> | undefined,
+      anchorCol: undefined as ViewerColumn | undefined,
       isCursorText: true,
       isCellTextSelectable: false,
       measureElement: vi.fn(),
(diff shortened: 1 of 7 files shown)


amorey:
@codex review

PR #1144, opened 2026-05-09
✨ Add bats tests and shellcheck for install script.

### LINKED ISSUE #1084: Refactor `install.sh` script and write unit tests for it using `bats`
Currently, our [`install.sh`](https://www.kubetail.com/install.sh) script isn't modularized and doesn't have any unit tests. We should import it into this library and write unit tests for it that get executed as part of our pre-merge checks:
https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/ci.yml

KaanSimsek:
Related issues:
Closes #1084

## Summary

The install script is refactored into small, testable functions, covered by bats tests that run in CI. The entry guard is fixed so curl … | bash works with set -u (no unsafe BASH_SOURCE[0] access). Optional KUBETAIL_REPO_URL and KUBETAIL_INSTALL_DIR support mirrors and non-root test installs. CI runs shellcheck on install.sh before bats. This addresses safer distribution and regression coverage for the documented one-liner install flow.

## Key Changes

* Add install.sh at the repo root: kubetail_* helpers (platform detection, download, checksum, install), kubetail_main, EXIT trap to clean temp artifacts on failure, stricter set -euo pipefail, curl with long flags, mkdir -p for Windows install dir, TMPDIR-aware temp paths.
* Main only runs when the file is executed, not when sourced: if [[ "${BASH_SOURCE:-$0}" == "${0}" ]]; then kubetail_main "$@".
* Add test/install/install.bats and test/install/test_helper.bash with mocked uname/curl to exercise all helpers and an end-to-end kubetail_main path without hitting the network.
* Add install-script-bats GitHub Actions job: shellcheck install.sh, install bats via apt, run bats test/install.

## Checks
I have tested on macOS, Windows, and Linux, and was able to successfully install Kubetail using this script on all three platforms:
curl -sS https://raw.githubusercontent.com/KaanSimsek/kubetail/chore/install-sh-bats-ci/install.sh | bash

## Checklist

- [ x ] Add the correct emoji to the PR title
- [ x ] Related issue linked above, if any
- [ x ] Commit messages use [conventional commit](https://www.conventionalcommits.org) format
- [ x ] Changes are minimal and focused

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,9 +111,80 @@ jobs:
         working-directory: ./dashboard-ui
         run: pnpm lint
 
+  install-script-shellcheck:
+    name: Install script · shellcheck
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install shellcheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+      - name: Run shellcheck
+        run: shellcheck scripts/install.sh
+
   ###########################
   # Test jobs
   ###########################
+
+  install-script-bats:
+    name: Install script · bats
+    needs:
+      - install-script-shellcheck
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - linux
+          - macos-15
+          - macos-26
+          - windows
+        arch:
+          - amd64
+          - arm64
+        include:
+          - os: linux
+            arch: amd64
+            runner: ubuntu-24.04
+          - os: linux
+            arch: arm64
+            runner: ubuntu-24.04-arm
+          - os: macos-15
+            arch: amd64
+            runner: macos-15-intel
+          - os: macos-15
+            arch: arm64
+            runner: macos-15
+          - os: macos-26
+            arch: arm64
+            runner: macos-26
+          - os: windows
+            arch: amd64
+            runner: windows-2025
+          - os: windows
+            arch: arm64
+            runner: windows-11-arm
+        exclude:
+          - os: macos-26
+            arch: amd64
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install bats (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y bats
+      - name: Install bats (macOS)
+        if: runner.os == 'macOS'
+        run: brew install bats-core
+      - name: Install bats (Windows)
+        if: runner.os == 'Windows'
+        shell: bash
+        run: npm install -g bats
+      - name: Run bats
+        run: bats scripts/
+        shell: bash
 
   go-test:
     name: Go · Test
--- a/scripts/install.bats
+++ b/scripts/install.bats
@@ -1,0 +1,368 @@
+#!/usr/bin/env bats
+# shellcheck disable=SC2031
+
+install_sh_source() {
+  KB_INSTALL_SH="$(cd "$(dirname "$BATS_TEST_FILENAME")/.." && pwd)/scripts/install.sh"
+  # shellcheck source=/dev/null
+  source "$KB_INSTALL_SH"
+}
+
+mock_uname_bin() {
+  local d="${1:?mock bin dir}"
+  mkdir -p "$d"
+  cat >"$d/uname" <<'EOF'
+#!/usr/bin/env sh
+case "${1-}" in
+  -s) printf '%s\n' "${UNAME_S:?}" ;;
+  -m) printf '%s\n' "${UNAME_M:?}" ;;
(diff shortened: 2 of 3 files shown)


KaanSimsek:
@codex review

amorey:
This is awesome! Thanks!

@codex review

amorey:
@KaanSimsek It's outside the scope of the original issue but do you want to implement the codex suggestion as well. We can treat it as a separate issue or roll it into this PR if you prefer.

amorey:
Nevermind it looks like you did it already 🙏

KaanSimsek:
> Nevermind it looks like you did it already 🙏

Thanks! Yes, it was a straightforward change so I went ahead and rolled it into this PR

scripts/install.sh
amorey, line 1: I think putting `install.sh` and `tests/` at the top-level could be confusing to new contributors. Instead, what do you think about putting the script and the test together in a new `scripts/` directory:
```
scripts/
  - install.sh
  - install.bats
```

scripts/install.bats
amorey, line 1: Can we include this code in `install.bats`?

install.sh
```diff
@@ -0,0 +1,162 @@
...
+  else
+    echo "🎉 Installation complete"
+  fi
+
+  echo "🚀 Have fun tailing your Kubernetes logs!"
+}
+
+if [[ "${BASH_SOURCE:-$0}" == "${0}" ]]; then
+  kubetail_main "$@"
+fi
```
amorey:
```suggestion
fi

```

.github/workflows/ci.yml
```diff
@@ -115,6 +115,20 @@ jobs:
   # Test jobs
   ###########################
 
+  install-script-bats:
+    name: Install script · bats
+    runs-on: ubuntu-24.04
```
amorey:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        os:
          - linux
          - macos-15
          - macos-26
          - windows
        arch:
          - amd64
          - arm64
        include:
          - os: linux
            arch: amd64
            runner: ubuntu-24.04
          - os: linux
            arch: arm64
            runner: ubuntu-24.04-arm
          - os: macos-15
            arch: amd64
            runner: macos-15-intel
          - os: macos-15
            arch: arm64
            runner: macos-15
          - os: macos-26
            arch: arm64
            runner: macos-26
          - os: windows
            arch: amd64
            runner: windows-2025
          - os: windows
            arch: arm64
            runner: windows-11-arm
        exclude:
          - os: macos-26
            arch: amd64
    runs-on: ${{ matrix.runner }}
```

amorey (changes requested):

scripts/install.sh
KaanSimsek: "Good point! Should I keep a tests/ subdirectory under scripts/ for the install.bats and test_helper.bats, or just put everything flat under scripts/?"

scripts/install.bats
KaanSimsek: Could you clarify which code you'd like included in install.bats?

amorey: I just meant the contents of `test_helper.sh`. It's not a big deal if it makes more sense to keep it separate, just curious.

amorey:
It looks like `shellcheck` isn't available on the macos and windows runners. Let's move that step into its own job in the "Pre-flight jobs" group and then add it as a dependency to the tester.

scripts/install.bats
KaanSimsek: Got it! I kept it separate to avoid making install.bats too large, but I went ahead and merged them into a single file.

KaanSimsek:
> It looks like `shellcheck` isn't available on the macos and windows runners. Let's move that step into its own job in the "Pre-flight jobs" group and then add it as a dependency to the tester.

Should I also exclude Windows from the bats matrix since bats is a bash testing framework and may not work reliably on Windows runners?

amorey:
> Should I also exclude Windows from the bats matrix since bats is a bash testing framework and may not work reliably on Windows runners?

Oh right, we need to install bats on Windows (and macOS). Check out how the CI job does it in our sister project here:
https://github.com/kubetail-org/kstack/blob/main/.github/workflows/ci.yml

The install script should work on Windows too so ideally we should include it in the bats matrix.

amorey:
Great! The tests passed! Can you squash the commits then I can merge?

amorey:
Thanks! Merged 🥳. Sorry, I just saw your last push. Unfortunately GitHub doesn't send reaction notifications so I didn't see your 👍 until just now.

KaanSimsek:
> Thanks! Merged 🥳. Sorry, I just saw your last push. Unfortunately GitHub doesn't send reaction notifications so I didn't see your 👍 until just now.

No worries, thanks for helping me contribute to this project. You are doing a wonderful job maintaining this project.

PR #1158, opened 2026-05-24
🎣 Tighten Sec-Fetch-Site CSRF check to reject same-site

amorey:
## Summary

Follow-up to #1155. The denylist installed in that PR was `["cross-site", "cross-origin"]`, but `cross-origin` isn't a valid `Sec-Fetch-Site` value per the Fetch Metadata spec — the real cross-origin values are `cross-site` and `same-site`. A `same-site` request originates from a sibling subdomain on the same eTLD+1 (e.g. `evil.example.com` → `dashboard.example.com`), which is still cross-origin in CSRF terms. The CSRF token remains the primary defense, but Layer 1 should actually catch this case instead of matching a value browsers never send.

## Key Changes

- Replace the dead `cross-origin` entry in `rejectedSecFetchSite` with `same-site` so sibling-subdomain requests are rejected outright.
- Update the dashboard middleware unit tests: move the `same-site` case from the Allows table to the Forbids table, drop the bogus `cross-origin` Forbids case, and add a `none` Allows case to keep coverage of the pass-through set.
- Update the e2e CSRF tests so `same-site` is in the rejected parametrization and the allowed set is `[missing, none, same-origin]`.

## Checklist

### DIFF
--- a/e2e/test_csrf.py
+++ b/e2e/test_csrf.py
@@ -138,22 +138,23 @@ async def _ws_send_init(
         return None
 
 
-# Explicitly cross-origin values that the CSRF gate must reject outright,
-# regardless of any accompanying token.
+# Cross-origin Sec-Fetch-Site values that the CSRF gate must reject outright,
+# regardless of any accompanying token. "same-site" covers sibling subdomains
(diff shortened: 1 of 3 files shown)


amorey:
@codex review

PR #1159, opened 2026-05-24
✨ Make PR checks diff-aware

### LINKED ISSUE #1154: Make PR checks diff-aware
Currently, we run all of our CI workflow jobs on every PR commit ([.github/workflows/ci.yml](https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/ci.yml)). Although we've done a lot of work to optimize the workflow, the checks still take a while to run and it would be better to make the jobs diff aware so we only run tests on the code that changed. It seems like [`dorny/paths-filter`](https://github.com/dorny/paths-filter) is a popular choice to do this so that would be a good starting point.

Tiago-Vier-Preto:
Closes #1154 

## Summary

This PR optimizes our GitHub Actions CI pipeline by making it diff-aware. Currently, every PR commit triggers all CI jobs across all languages and components, which is time-consuming. By introducing `dorny/paths-filter`, we can detect which files changed and conditionally skip jobs for untouched components (e.g., skipping Rust tests if only Go files were modified). This will provide faster feedback loops for contributors and save CI runner minutes.

## Key Changes

- Added a new `changes` job at the start of the workflow using `dorny/paths-filter` to detect file changes across `go`, `rust`, `typescript`, `shell`, `python`, `docker`, and `workflow` paths.
- Updated pre-flight jobs (lint, vet, vendor) with `needs: changes` and `if` conditions to run only when relevant files are modified.
- Updated test, build, and downstream jobs with `needs: changes` and conditional `if: always() && !failure() && !cancelled() && (...)` logic so they safely bypass intentionally skipped upstream dependencies.
- Added fallbacks to ensure all jobs still run unconditionally if the workflow is triggered manually (`workflow_dispatch`) or if the CI workflow files themselves are modified.

## Checklist

- [x] Commit messages use conventional commit format

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,57 @@ on:
 
 jobs:
   ##############################
+  # Changes detection
+  ##############################
+
+  changes:
+    name: Detect Changes
+    runs-on: ubuntu-24.04
+    outputs:
+      go: ${{ steps.filter.outputs.go }}
+      rust: ${{ steps.filter.outputs.rust }}
+      typescript: ${{ steps.filter.outputs.typescript }}
+      shell: ${{ steps.filter.outputs.shell }}
+      python: ${{ steps.filter.outputs.python }}
+      docker: ${{ steps.filter.outputs.docker }}
+      workflow: ${{ steps.filter.outputs.workflow }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            go:
+              - 'modules/**'
+              - 'go.work'
+              - 'go.work.sum'
+              - 'Makefile'
+            rust:
+              - 'crates/**'
+              - 'Cargo.toml'
+              - 'Cargo.lock'
+              - 'Makefile'
+            typescript:
+              - 'dashboard-ui/**'
+              - 'Makefile'
+            shell:
+              - 'scripts/**'
+            python:
+              - 'e2e/**'
+            docker:
+              - 'build/package/**'
+            workflow:
+              - '.github/workflows/ci.yml'
+              - '.github/actions/**'
+
+  ##############################
   # Pre-flight jobs
   ##############################
 
   go-lint:
     name: Go · Lint
+    needs: changes
+    if: github.event_name == 'workflow_dispatch' || needs.changes.outputs.go == 'true' || needs.changes.outputs.workflow == 'true'
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -10,12 +10,58 @@ on:
...
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            go:
+              - 'modules/**'
+              - 'go.work'
+              - 'go.work.sum'
```
amorey: These files are in `modules/`, does `modules/**` cover them?

.github/workflows/ci.yml
```diff
@@ -10,12 +10,58 @@ on:
...
+              - 'Makefile'
+            rust:
+              - 'crates/**'
+              - 'Cargo.toml'
+              - 'Cargo.lock'
```
amorey: These files are in `crates/`, does `crates/**` cover them?

.github/workflows/ci.yml
```diff
@@ -10,12 +10,58 @@ on:
...
+              - 'Makefile'
+            shell:
+              - 'scripts/**'
+            python:
+              - 'e2e/**'
+            docker:
+              - 'build/package/**'
+            workflow:
+              - '.github/workflows/ci.yml'
+              - '.github/actions/**'
```
amorey: Let's ignore these changes for now and just focus on changes in the core component languages (Go, Rust, TypeScript)

amorey (changes requested):
Thanks! I added some comments inline about specific files (e.g. `go.work`, `Cargo.toml`) and also about non-core changes (e.g. `shell`, `python`). In addition, here are some other suggestions:

- I think we can remove `Makefile` from the filters for now. There are some edge cases where we would want it but we can deal with that later.
- In general, I think we can simplify things by only checking for changes in the core languages: Go, Rust and TypeScript
- I think we can remove the `always()`, `!failure()`, `!cancelled()` calls in the if statements
- Instead of checking for `github.event_name == 'workflow_dispatch'` what would be really nice is to be able to override the `dorny/paths-filter` output and set all values to true when we want to run the full suite. Here's an example of how to set `workflow_dispatch` options: https://github.com/kubetail-org/kubetail/blob/main/.github/workflows/release-cli.yml
- It would also be convenient to have an `any` output that we can use to run the docker builds and e2e tests if there's a change in any core language

amorey:
Also, I just noticed that you did a merge commit from `main` into your branch. To keep the history linear, can you undo that commit and do a rebase instead? For next time, you can even leave your branch out-of-date and I can do the rebase. If there are no conflicts it's easy to do it on my side so I don't think it's worth it to place the burden on contributors.

amorey:
Thanks! This is great. It's going to speed up our PR checks dramatically.

Issue #999, opened 2026-03-04
`permission denied: "list pods/log" in namespace "all"` when using dashboard with authMode: token.


NicholasPini:
I have installed on my kubernetes cluster Kubetail 0.18.1, using Helm. I have set `authMode: token` because I want to eventually make the Kubetail dashboard available on the internet, and I need password protection. I have created the following serviceAccount: 

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: kubetail-viewer
  namespace: kubetail
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kubetail-viewer
rules:
  - apiGroups: [""]
    resources: [namespaces, nodes]
    verbs: [get, list, watch]
  - apiGroups: ["", apps, batch]
    resources:
      - cronjobs
      - daemonsets
      - deployments
      - jobs
      - pods
      - pods/log
      - replicasets
      - statefulsets
    verbs: [get, list, watch]
  - apiGroups: [authentication.k8s.io]
    resources: [tokenreviews]
    verbs: [create]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: kubetail-viewer
subjects:
  - kind: ServiceAccount
    name: kubetail-viewer
    namespace: kubetail
roleRef:
  kind: ClusterRole
  name: kubetail-viewer
  apiGroup: rbac.authorization.k8s.io
```

If I now generate a token from this service account and access the dashboard, I see the list of pods and other resources, but I get this error in the bottom right:
```
[GraphQL Error] kubernetes://kubetail-helm-cluster-agent:50051/: rpc error: code = PermissionDenied desc = permission denied: `list pods/log` in namespace `all`
```

I also cannot see any logs. If I run `kubectl auth can-i list pods/log --all-namespaces --as=system:serviceaccount:kubetail:kubetail-viewer`, I get `yes` as a response.

What am I doing wrong?

amorey:
@NicholasPini Thanks again for letting us know about the problem. It was a bug on our end and should be fixed in the latest helm chart (kubetail-0.18.2). This manifest should work for the `kubetail-viewer` serviceaccount:
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: kubetail-viewer
  namespace: kubetail
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kubetail-viewer
[...]

Issue #220, opened 2025-04-27
kubetail serve fails with "at the cluster scope" errors due to service account permission level


mark-dessain-maersk:
For context, we have a company wide cluster whereby I am part of an application team with access to a single namespace. I have no permissions over any cluster wide settings or "all namespaces".

My service account only has access to list all name spaces and to run commands on my own namespace (in this case it is called "orbit").

Such that the following kubectl commands work as shown:

```
➜ kubectl get namespaces
NAME                        STATUS   AGE
...                         Active   627d
orbit                       Active   2y310d
abc                         Active   635d
...                         Active   635d

➜ kubectl get pods --all-namespaces
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:serviceaccounts:orbit-ci-sa" cannot list resource "pods" in API group "" at the cluster scope

➜ kubectl get pods --namespace abc
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:serviceaccounts:orbit-ci-sa" cannot list resource "pods" in API group "" in the namespace "abc"

➜ kubectl get pods --namespace orbit
NAME                                   READY   STATUS      RESTARTS      AGE
community-6d7695f448-vpm8d             1/1     Running     0             51d
...
```

The `kukubetail logs --follow 'deployments/*'` works perfectly fine but when I try and use the dashboard ` kubetail serve` it fails to show anything.

The dashboard lists the sidebar counts but all other graphql calls fail.

<img width="1045" alt="Image" src="https://github.com/user-attachments/assets/6b5e74f2-a7fe-4a14-9ca2-2171625417bf" />

<img width="1045" alt="Image" src="https://github.com/user-attachments/assets/f00f1a36-553c-4421-a0bb-13e363389b12" />

---

Looking at the network request I can see that the requests for the sidebar which work include the namespace, where as the requests for the actual middle data do not include the namespace.

I assume this combined with my namespace only permissions causes it to fail.

<img width="916" alt="Image" src="https://github.com/user-attachments/assets/bcbd2609-9f39-430b-8fa2-5b4d0a1ccaed" />

The counters show due to this api call passing in the namespace - https://github.com/kubetail-org/kubetail/blob/main/dashboard-ui/src/pages/home.tsx#L811

<img width="930" alt="Image" src="https://github.com/user-attachments/assets/19426562-2f59-4e40-8dc7-1909fb2fc9b9" />

[...]

Issue #21, opened 2024-02-28
namespaced installation supported?


jkroepke:
Hi,

I would like to ask, if kubetail supports a namespaced installation as well, e.g. auth-mode=cluster and namespaced Rolebindings only.

Thanks!

amorey:
The helm chart doesn't currently support namespaced rolebindings but we could add it there. In addition, there's an undocumented, experimental option in the kubetail named `namespace` that restricts access to a given namespace at the UI-level. It's currently being used on the demo site (in addition to rbac under the hood): https://www.kubetail.com/demo.

Would that work for you?

Also - thanks for the helm pull request! If you're interested in adding RoleBinding support too, that would be very much appreciated otherwise I'll work on it asap.

jkroepke:
> Also - thanks for the helm pull request! If you're interested in adding RoleBinding support too, that would be very much appreciated otherwise I'll work on it asap.

Thats something I can cover. There are also a few options missing in helm chart that I would like to add, e.g. `securityContext`,`autoMountServiceAccountToken` and or tpl support at ingress. I have already a commit in draft.

amorey:
Sounds great! Looking forward to checking out the PR.

jkroepke:
> Would that work for you?

I had no success here:

I'm using the image `kubetail/kubetail:0.1.9` and I have assign RoleBindings and a Role to a service account.

![image](https://github.com/kubetail-org/kubetail/assets/1560587/5f5d7e92-a047-4dee-8f8b-3018930695fb)

The logs remains empty and on browser console, there is an error visible that the serviceAccount try to access nodes (which are a cluster-scoped resource). 

Maybe there can be implement a try/catch we ignores the error, if the nodes are not available.

amorey:
Currently you have to give it cluster-scoped access to nodes and namespaces:
```yaml
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: kubetail
rules:
- apiGroups: [""]
  resources: ["nodes", "namespaces"]
  verbs: ["get", "list", "watch"]
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: kubetail
subjects:
- kind: ServiceAccount
  namespace: kubetail
  name: kubetail
roleRef:
  kind: ClusterRole
  name: kubetail
  apiGroup: rbac.authorization.k8s.io
---
kind: Role
[...]

PR #1175, opened 2026-07-22
✨ Support namespace-scoped tenants on a shared dashboard (auth-mode: token)

krim404:
## Summary

When running the dashboard with `auth-mode: token`, a session always operates at global/cluster scope. For a user whose token only has namespace-scoped RBAC (no cluster-wide access), that means the dashboard tries to list everything and the UI fills up with "Forbidden ... at the cluster scope" errors, because it never narrows down to the namespaces the token is actually allowed in (see #220 and #999, and the discussion in #21).

This PR makes the dashboard tenant-capable: one shared instance can serve users that are each restricted to their own namespace(s), instead of installing a separate kubetail per namespace/tenant with its own RBAC. A login can declare which namespace(s) its session is scoped to, so the queries match the token's RBAC and the UI stays clean.

This is a session scope, not access control: the token still needs matching Kubernetes RBAC for the selected namespaces, nothing here grants anything.

## Key Changes

- New config option `dashboard.allow-namespace-override` (default `false`), so nothing changes unless it is explicitly enabled.
- The login accepts an optional repeated `namespaces` field. The values are stored in the session and applied to all queries, subscriptions and log downloads for that session.
- The override can only shrink `allowed-namespaces`, never widen it. If the static list is set, a login requesting a namespace outside of it gets a 422. If the config changes after login, the intersection is re-applied per request on the server side.
- `/api/auth/session` reports the effective lock as `namespace_lock`, and the namespace pickers reflect it ("All allowed namespaces" plus a tooltip).
- Tested with `go test ./...` on the dashboard and shared modules, plus tsc/eslint for the UI changes.

## Usage / recommended setup

kubetail itself has no general multi-user login (username/password, etc.), and that is intentionally out of scope for this PR. On its own the new `namespaces` field is only a plumbing hook: there is no login-page UI element for it. To be useful for tenants you put an authenticating layer in front (nginx auth, oauth2-proxy, or any reverse proxy) that:

1. authenticates the user (e.g. username/password per tenant, or OIDC),
2. knows which namespace(s) that user is allowed to see, and
3. performs the token login on their behalf, including the `namespaces` field.

Example with an OIDC flow (this is how I run it):

- oauth2-proxy sits in front of the dashboard and authenticates users against an OIDC provider (Authentik in my case),
- the user's namespace assignment comes from their identity (a per-user claim / group), and is passed as `namespaces` when the session is established,
- the dashboard then scopes that session to those namespaces, and the pod is reached with the user's own bearer token so Kubernetes RBAC still enforces access.

The result is a single dashboard behind the proxy where each user only sees (and is only allowed to see) their own namespace, without granting anyone cluster-wide visibility and without a separate deployment per tenant.

A built-in namespace picker at login time (so a user could choose without a fronting proxy) would be a natural follow-up, but is deliberately not part of this PR.

## Checklist


krim404:
Applied the suggested fix for the cluster-api proxy path: the session namespace scope is now forwarded to the cluster-api as an `X-Forwarded-Namespaces` header (mirroring the existing `X-Forwarded-CSRF-Token` handling), the dashboard proxy strips any client-supplied copy, and the cluster-api narrows its allowed-namespaces from it in the same way the dashboard does. It can only shrink the allow-list, never widen it.

Heads up: I can't test this path end to end myself since I don't run the cluster-api component (it's disabled in my setup), so that part is covered only by unit tests, not a live run.

As i said, without testing i relied on the help of Fable 5.

PR #1181, opened 2026-09-16
Pin actions/checkout to commit SHA in CI workflows

sapienfrom2000s:
## Summary

Pin `actions/checkout@v6` to its resolved commit SHA
(`d23441a48e516b6c34aea4fa41551a30e30af803`) in `ci.yml` and
`post-merge-checks-riscv64.yml` to guard against tag mutation/supply-chain
risk in CI.

## Key Changes

- Replace all `actions/checkout@v6` references with
  `actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6`
  (21 occurrences across `.github/workflows/ci.yml` and
  `.github/workflows/post-merge-checks-riscv64.yml`)

## Checklist

- [x] Add the correct emoji to the PR title
- [ ] Related issue linked above, if any
- [x] Commit messages use [conventional commit](https://www.conventionalcommits.org) format
- [x] Changes are minimal and focused

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1182, opened 2026-09-17
Add Renovate to pin GitHub Actions to SHA

sapienfrom2000s:
## Summary

Sets up Renovate, scoped to only the `github-actions` manager, to keep
`uses:` references in workflow files pinned to a commit SHA (with a
version comment) instead of a mutable tag. Complements #1181, which
did this pinning manually for `actions/checkout@v6`.

## Key Changes

- `.github/renovate.json` — enables Renovate for the `github-actions`
  manager only (no npm/Go/Rust dependency updates), extending
  `helpers:pinGitHubActionDigestsToSemver` so digest pins are tracked
  with proper major/minor/patch classification instead of opaque
  digest-only bumps.
- `.github/workflows/renovate.yml` — runs Renovate itself via
  `renovatebot/github-action` on a GitHub-hosted runner (daily cron +
  manual dispatch), rather than via npm locally. Both `uses:` steps in
  this workflow are pinned to SHA.

## Notes for reviewers

- Requires a `RENOVATE_TOKEN` repo secret (PAT or GitHub App
  installation token with `contents: write` + `pull-requests: write`)
  to be created before this can open PRs — the default `GITHUB_TOKEN`
  can't trigger `ci.yml` on PRs it opens.
- Once merged and the token is set, the scan is repo-wide: it will
  also propose pinning the remaining `v5`/`v4` action refs elsewhere
  in `.github/workflows/` (not just the ones already pinned in #1181).

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


amorey:
Nice! Some suggestions:
- `renovate.json` should go in `.github/ci-config/`
- Instead of running on a cronjob, let's run it on pushes into `main`

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,28 @@
...
+permissions:
+  contents: read
+
+concurrency:
+  group: renovate
+  cancel-in-progress: true
+
+jobs:
+  renovate:
+    runs-on: ubuntu-24.04
```
amorey, line 18:
```suggestion
    runs-on: ubuntu-24.04
    environment: production
```

amorey (changes requested):

amorey (approved):

